PR #405, opened 2012-10-12
Added Composer compatibility.

Blackskyliner:
Original composer.json is taken from helios-ag's fork. I just edited it back to your name.

This will add Composer compatibility.

Maybe you could also add this Project, after the Pull of the composer.json, to https://packagist.org, so there would be always an upstream version and no maintaining from helios-ag would be needed. 

If you then create the Packagist Git Hook on this repo your updates will automatically propagated.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -1,0 +1,18 @@
+{
+    "name": "studio-42/fm-elfinder",
+    "description": "File Manager for Web",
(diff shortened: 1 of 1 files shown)


nao-pon:
Thanks! #929

Issue #136, opened 2011-09-24
Error in elFinderVolumeDriver (PHP strict standards)


jaygridley:
Getting this when loading elFinder...

PHP Strict standards: Only variables should be passed by reference in /hosting/www/.../elFinder/elFinderVolumeDriver.php:642

```
which is && preg_match($regexp, array_shift(explode(';', @finfo_file(finfo_open(FILEINFO_MIME), __FILE__))))) {
```

should be fixed I think.

PR #416, opened 2012-10-22
Fix for issue 136

dominionit:
https://github.com/Studio-42/elFinder/issues/136
    
    Problem on older php systems < 5.3 seems that MAGIC variable is not
    set and even if you pass NULL or not pass any value as second variable it
    does not read from MAGIC, thus passing the MAGIC enviroment variable as
    second paramater should be acceptable for older and newer php systems.


nao-pon:
#350 and  da50bbad3912a6303ad3730fe9df601d9e1989b2

PR #418, opened 2012-10-23
 Opening directory-hang (session_start hang bug) - After failed ajax call attempt

dominionit:
Opening directory-hang (session_start hang bug) - After failed ajax call attempt
    
    This seems to solve the issue where if elFinder had a failed ajax call to
    for example connector.php and crashes, that it does not hang on the
    Opening Directory progress bar.
    
    Actually the session_start hangs and you need to delete your cookie php session id
    to get it to work again, where this will make it work correctly it seems
    after some testing


nao-pon:
I think the PHP version and accelerator influence this problem. Reinspection, please.

PR #545, opened 2013-03-27
Added bower support

eymengunay:
http://twitter.github.com/bower


fcaivano:
Why is it closed? No bower support was added ever. Is there no intention in doing so?

nao-pon:
@fcaivano Thanks! I added bower.json @62e73c4857e7fc7ceccbe01d1e822109840ae0a1

PR #682, opened 2013-09-11
Add fileinfo extension required

ifrin:


nao-pon:
`finfo` is recommendation.

PR #752, opened 2013-12-18
New option soundPath

vagnerleitte:
Especify a custom path to the rm.wav sound.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -192,6 +192,14 @@ window.elFinder = function(node, opts) {
 		 * @default 400
 		 **/
 		height = 400,
+		
+		/**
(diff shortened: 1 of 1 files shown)


nao-pon:
@vagnerleitte Thanks! It will marge to 2.1. :+1:

ichuot:
I think we can get full url of elfinder for default value.
```javascript
var ELFINDER_URL = (function() {
  var script_tags = document.getElementsByTagName('script');
  var url = script_tags[script_tags.length-1].src;
  url = url.substring(0, url.lastIndexOf("/"));
  var to = url.lastIndexOf('/');
  to = to == -1 ? url.length : to + 1;
  return url.substring(0, to);
})();
```

PR #795, opened 2014-02-20
Fix Windows style exploding root path

actorius:
If root path look like:
_C:\xampp\htdocs\frontend/../../backend/www_
code generate error: Root folder does not exist. 
Because 
```
$comps = explode('/', $path);
```
explode only unix separators, later in foreach it popup full first part _C:\xampp\htdocs\frontend._
In the end return only **../backend/www**

I fix this problem by replacing all windows separators **\** to unix, and other code work fine.

### DIFF
--- a/php/elFinderVolumeLocalFileSystem.class.php
+++ b/php/elFinderVolumeLocalFileSystem.class.php
@@ -170,7 +170,9 @@ class elFinderVolumeLocalFileSystem extends elFinderVolumeDriver {
 		}
 			
 		$initial_slashes = (int) $initial_slashes;
-
+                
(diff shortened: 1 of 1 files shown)


php/elFinderVolumeLocalFileSystem.class.php
```diff
@@ -170,7 +170,9 @@ protected function _normpath($path) {
 		}
 			
 		$initial_slashes = (int) $initial_slashes;
-
+                
+                $path = str_replace(array('/', '\\'), '/', $path);
```
Rillke: Tabs have been used for indention in this file.

nao-pon:
May be It fixed on 2.1.0. Please check on 2.1.x.

PR #834, opened 2014-04-15
2.1

farazilu:
added js and css ini to development versions.


nao-pon:
The absolute path of URL isn't general-purpose. And Please create new branch for pull request.

PR #860, opened 2014-05-29
Fix input color

adisos:
I don't know why, but my input has background and font color white. In dev tools I couldn't point to input to saw css rules, because when I focus out from input it disappears.


Rillke:
Perhaps some CSS originating from the surrounding page interfering?

nao-pon:
In accordance with the respective environment, please coordinate your CSS.

PR #870, opened 2014-06-26
add md5sum in commands.info

gabrielrota:
Enabled md5sum value for file into info dialog.
Disabled by default.
Example for enabling:
$('#finder').elfinder({ commandsOptions : { info : { md5sum : true } } });

### DIFF
--- a/php/elFinderVolumeDriver.class.php
+++ b/php/elFinderVolumeDriver.class.php
@@ -1182,6 +1182,22 @@ abstract class elFinderVolumeDriver {
 	}
 	
 	/**
+	 * md5sum from file
+	 *
+	 * @param  string   file hash
+	 * @return string
+	 * @author gabriel.rota@gmail.com
+	 **/
+	public function md5sum($hash) {
+		if (($file = $this->file($hash)) == false
+		|| $file['mime'] == 'directory') {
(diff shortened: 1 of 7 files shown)


php/elFinderVolumeDriver.class.php
```diff
@@ -1182,6 +1182,22 @@ public function open($hash) {
...
+	 * @return string
+	 * @author gabriel.rota@gmail.com
+	 **/
+	public function md5sum($hash) {
+		if (($file = $this->file($hash)) == false
+		|| $file['mime'] == 'directory') {
+			return false;
+		}
+		
+		return $this->_md5sum($this->decode($hash), 'rb');
```
Rillke: - What is the second argument for?
- Won't this explode for file systems other than the local FS? I suggest adding a stub or abstract method `elFinderVolumeDriver::_md5sum()`.

nao-pon:
On elFinder 2.1.x, we can now add any item to the info dialog. Take advantage of its features.

- https://github.com/Studio-42/elFinder/blob/140054e12682c9f7fbc16f8d33ff91b824002c32/js/elFinder.options.js#L220

Connector side should be its own override.

PR #887, opened 2014-08-19
regression from f38a3787 : we should use variable url 

anwfr:
regression from f38a3787 : we should use variable url which was set a few lines upper. 
Otherwise elfinder always opens files through connector 'file' command, even if we want to open files directly from remote URL.


nao-pon:
There is a possibility that a custom parameter of session ID etc are included in URL. Therefore I think URL should be hidden.
You can find Item's URL at `Link:` in Info dialog.

PR #928, opened 2014-12-02
[WIP] Add Flysystem driver

barryvdh:
This is a Volume Driver for Flysystem (https://github.com/thephpleague/flysystem)
This makes it easy to use Local, (S)FTP, AWS, Dropbox, etc without writing custom volume drivers..

This is tested against the latest `dev-master`, but a stable 1.x release should come soon. Probably best to wait before the final changes.

TODO:
 - [x] Check all adapters (Tested: Local, FTP, SFTP, Dropbox)
 - [x] Think about best practice, should a FileSystem instance be passed, or just the Adapter? <- Both
 - [x] Simplify the path: ~~is the path needed, or just use the root from the adapter?~~ Yes and already works.
 - [ ] Wait for final (1.x) release

Example usage: pass a Flysystem instance to the options. Path can be '/' to use the root, or a relative path.

    use League\Flysystem\Filesystem;
    use League\Flysystem\Adapter\LocalAdapter;
    use League\Flysystem\Adapter\SftpAdapter;
        
    // Local Adapter
    array(
        'driver' => 'Flysystem', 
        'path' => '/',
        'URL' => asset('content'), 
        'filesystem' => new Filesystem(new LocalAdapter('/var/www/html/content'))
    ),
    
    // SFTP Adapter
    array(
        'driver' => 'Flysystem',
        'path' => 'content',
        'URL' => asset('content'),
        'filesystem' => new Filesystem(new SftpAdapter(
                [
                    'host' => 'localhost',
                    'username' => 'username',
                    'password' => '123',
                    'root' => '/var/www/html',
                ]
            )),
    ),

Any thoughts?


nao-pon:
Functions `root`, `normpath`, `relpath`, `abspath` and etc. That was implemented before I participated. So I have to understand correctly about that functions. So I'll think it over.

nao-pon:
And Happy New Year!　:smile:

barryvdh:
I've added a method to resize the the images also. 
But I feel like this is probably a lot the same for all non-local volume drivers. I just used the _getContents and _filePutContents method to read/write to a temporary file, so it should be the same for all drivers..
Perhaps it's worth to put it the elfinderVolumeDriver and just let the local driver use the 'native' methods?

nao-pon:
> Perhaps it's worth to put it the elfinderVolumeDriver and just let the local driver use the 'native' methods?

I think so, too. I'm thinking as follows.

- elFinderVolumeDriver
```php
	public function resize($hash, $width, $height, $x, $y, $mode = 'resize', $bg = '', $degree = 0) {
		$this->lastResizeInfo = array(
			'path' => '',
			'size' => array()
		);
		
		if ($this->commandDisabled('resize')) {
			return $this->setError(elFinder::ERROR_PERM_DENIED);
		}
		
		if (($file = $this->file($hash)) == false) {
			return $this->setError(elFinder::ERROR_FILE_NOT_FOUND);
		}
		
		if (!$file['write'] || !$file['read']) {
			return $this->setError(elFinder::ERROR_PERM_DENIED);
		}
		
		$path4stat = $this->decode($hash);
		if (method_exists($this, 'getHandyFile')) {
			if (!$path = $this->getHandyFile($path4stat)) {
				return false;
			}
		} else {
			$path = $path4stat;
		}
		$this->lastResizeInfo['path'] = $path;
		
		if (!$this->canResize($path, $file)) {
			return $this->setError(elFinder::ERROR_UNSUPPORT_TYPE);
		}

		switch($mode) {
			
			case 'propresize':
				$result = $this->imgResize($path, $width, $height, true, true);
				break;

			case 'crop':
				$result = $this->imgCrop($path, $width, $height, $x, $y);
				break;

			case 'fitsquare':
				$result = $this->imgSquareFit($path, $width, $height, 'center', 'middle', ($bg ? $bg : $this->options['tmbBgColor']));
				break;

			case 'rotate':
				$result = $this->imgRotate($path, $degree, ($bg ? $bg : $this->options['tmbBgColor']));
				break;

			default:
				$result = $this->imgResize($path, $width, $height, false, true);
				break;
		}

		if ($result) {
			clearstatcache();
			$this->lastResizeInfo['size'] = @getimagesize($path);
			
			$this->rmTmb($file);
			$this->clearcache();
			return $this->stat($path4stat);
		}
		
		return false;
	}
```
ex. elFinderVolumeDropbox
```php
	protected function getHandyFile($path) {
		if ($handy = $this->getLocalName($path)) {
			if ($contents = $this->_getContents($path) and @file_put_contents($handy, $contents, LOCK_EX)) {
				return $handy;
			}
		}
		return false;
	}

	public function resize($hash, $width, $height, $x, $y, $mode = 'resize', $bg = '', $degree = 0) {
		$res = parent::resize($hash, $width, $height, $x, $y, $mode, $bg, $degree);
		$handy = $this->lastResizeInfo['path'];
		if ($res) {
			$path = $this->decode($hash);
			if ($fp = @fopen($handy, 'rb')) {
				if (!$this->_save($fp, $path, '', array())) {
					$res = false;
				}
				@fclose($fp);
				if ($res && $this->lastResizeInfo['size']) {
					$raw = $this->getDBdat($path);
					$raw['width'] = $this->lastResizeInfo['size'][0];
					$raw['height'] = $this->lastResizeInfo['size'][1];
					$this->updateDBdat($path, $raw);
				}
			} else {
				$res = false;
			}
		}
		$handy && file_exists($handy) && @unlink($handy);
		return $res;
	}
```

barryvdh:
Could probably be a getLocalFile() function also on the AbstractDriver, because it's always the same. And the local driver could just return the same filename, which we could check

    protected function getLocalFile($path)
    {
        $local_path = tempnam(sys_get_temp_dir(), 'elfinder');
        if (!$local_path) {
            return false;
        }

        $contents = $this->_getContents($path);
        if (!$contents) {
            return false;
        }
        
        $written = @file_put_contents($local_path, $this->_getContents($path), LOCK_EX);
        if ($written !== false) {
            return $local_path;
        }
    }

Check before deleting:

    $path = $this->decode($hash);
    $local_path = $this->getLocalFile($path);
    if (!$local_path) {
        $this->setError(elFinder::ERROR_FTP_DOWNLOAD_FILE, $local_path);
        return false;
    }
    //.....skip this part..
    if ($result) {
        if($local_path !== $path) {
            $contents = file_get_contents($local_path);
            if ($contents && !$this->_filePutContents($path, $contents)) {
                $this->setError(elFinder::ERROR_FTP_UPLOAD_FILE, $path);
                $this->deleteFile($local_path); //cleanup
            }
        }
        $this->clearcache();
        return $this->stat($path);
    }

So local just has to be:

    protected function getLocalFile($path)
    {
        return $path;
    }

But perhaps this is better fitted in a different issue.

nao-pon:
Thanks! that right. :+1:

barryvdh:
Re-submitted in #967 because squashing didn't work out..

barryvdh:
Ah cool you worked that out :)

I've create a seperate repository for the Flysystem driver, so it's easier to restrict to certain versions etc. (And Composer is required anyways)
https://github.com/barryvdh/elfinder-flysystem-driver

It also supports http://glide.thephpleague.com/ to create thumbnails or secure urls :)

PR #967, opened 2015-01-22
Implement Flysystem Volume Driver

barryvdh:
Replacement for #928 

I think this is good to go, at least for now. Flysystem just tagged 1.0.0 so should be stable :)


barryvdh:
Perhaps this would be better fit as a custom package, to take care of package etc. Let me think about it ;)

nao-pon:
I have just thought I'll merge. If you'd support continuously, please reopen. :-)

barryvdh:
Yeah I'm not sure, if it's a seperate package I can require dependancies and perhaps support new releases etc. And I can make updates/changes faster.
So perhaps better to keep it a seperate package and just list it in the readme?

nao-pon:
OK! Please send pull request about "README.md".;-)

abdala:
Just for reference:
@barryvdh: "I've create a seperate repository for the Flysystem driver, so it's easier to restrict to certain versions etc. (And Composer is required anyways)
https://github.com/barryvdh/elfinder-flysystem-driver"

PR #968, opened 2015-01-23
Remove invalid comments from CSS

barryvdh:
These `//` comments are not allowed in CSS, cause the jake build to fail.


nao-pon:
Thanks!

PR #969, opened 2015-01-23
Check if working path is normal path before delete

barryvdh:
$work_path could also be false, when creating failed.
Don't delete when the local file is the original file.


nao-pon:
:+1:

PR #996, opened 2015-03-10
FIX Message "Invalid backend response. Data is empty."

andersonscr:
json_encode needs it's input data to be encoded in UTF-8, so we added a an array walk to convert to UTF-8 any string that is not in that encoding yet.


nao-pon:
`utf8_encode` supported only ISO-8859-1 so it isn't possible to accept this request.

On 2.1 branch has Volume root options ` encoding` and `locale`. Please try 2.1 with these options.

- https://github.com/Studio-42/elFinder/wiki/Connector-configuration-options-2.1#encoding

PR #1002, opened 2015-03-28
Fix edge case when the root is / and no alias is provided.

therevmj:
While playing around I was having problems when I set the root to '/' and tracked it down to the rootName being set to '' due to the fact that basename (unsurprisingly) returns '' for '/'.  I think this is the best place for this change but it could be pulled into _basename in elFinderVolumeLocalFileSystem.class.php.  But that might break otherthings potentially.


nao-pon:
@therevmj Thanks!

PR #1003, opened 2015-03-29
fix _inpath when root is /

therevmj:
When / is the root, the '_inpath' function would compare /foo to // and thus fail.  This change detects if the path ends in a / and does not add an addition / in that situation.

I also added the min and full .js and .css files that are created during the build process to the .gitignore file.


nao-pon:
@therevmj Thanks for your PR. I fixed another solution. (553818690487b7ffebd33f1ec12c80529d45e09b)

PR #1009, opened 2015-04-13
2.1

standus:
Hi,
 
I would like to have ".quarantine" on one global place, so I changed code to:
$this->options['quarantine'] = 'tmp/.quarantine';
but now is problem that I have many many errors mkdir(), because elfinder class want to create directory in /image folder and not in my defined web root folder..
I found that it want create folder which is defined: $this->quarantine = $this->root.DIRECTORY_SEPARATOR.$this->options['quarantine'];
And that is the problem because I dont want it create in /images folder.
I can change it myself what I need, but with next update I must be very carefull and compare future changes with my improvement.
Is it possible on your site make next configuration for ".quarantine" folder? Something like : quarantineURL ?
I have similar specific configuration folder for tmbURL.
 
Thank you very much
Stan


nao-pon:
Please create new branch for pull request.

PR #1026, opened 2015-06-01
return ownership (uid/gid) and mode (permissions)

captin411:
I'm not sure what the process is to add information to the file structure returned when listing folders.

Ultimately, I would love to add in ownership and permission information to the structure returned when getting file listings -- just in the connector, not expecting it in the UI.

I also added a 'chmod' command which can be called like so:

connector.php?cmd=chmod&target=HASH&mode=0755

This will error out if the 'chmod' command is disabled, or if the effective UID of the process does not have permission to perform the chmod, or if the file/directory doesn't exist.

It returns a fresh stat of the file/directory if successful.

Did I go about this the right way?  If not, any suggestions or concerns with this approach?

```
{
    "name"   : "Images",             // (String) name of file/dir. Required
    "hash"   : "l0_SW1hZ2Vz",        // (String) hash of current file/dir path, first symbol must be letter, symbols before _underline_ - volume id, Required. 
    "phash"  : "l0_Lw",              // (String) hash of parent directory. Required except roots dirs.
    "mime"   : "directory",          // (String) mime type. Required.
    "ts"     : 1334163643,           // (Number) file modification time in unix timestamp. Required.
    "date"   : "30 Jan 2010 14:25",  // (String) last modification time (mime). Depricated but yet supported. Use ts instead.
    "size"   : 12345,                // (Number) file size in bytes
    "dirs"   : 1,                    // (Number) Only for directories. Marks if directory has child directories inside it. 0 (or not set) - no, 1 - yes. Do not need to calculate amount.
    "read"   : 1,                    // (Number) is readable
    "write"  : 1,                    // (Number) is writable
    "locked" : 0,                    // (Number) is file locked. If locked that object cannot be deleted,  renamed or moved
    "tmb"    : 'bac0d45b625f8d4633435ffbd52ca495.png' // (String) Only for images. Thumbnail file name, if file do not have thumbnail yet, but it can be generated than it must have value "1"
    "alias"  : "files/images",       // (String) For symlinks only. Symlink target path.
    "thash"  : "l1_c2NhbnMy",        // (String) For symlinks only. Symlink target hash.
    "dim"    : "640x480",            // (String) For images - file dimensions. Optionally.
    "volumeid" : "l1_",              // (String) Volume id. For root dir only.

    // my additions to LocalVolume driver

    "chmod" : 1,                      // (Number) will chmod work on this file or folder?
    "perms" : {
        "umask" : "0022",            // (String) the umask for the current running process
        "octal" : "0700",            // (String) what you would expect to pass to 'chmod'
        "mode"  : 16832,             // (Number) raw mode from stat or lstat call
    },
    "ownership" : {
        "owner"     : 1234,          // (Number) uid of file from stat or lstat call
        "ownerName" : "foobar"       // (String) the posix user name for that uid
        "group"     : 1234,          // (Number) gid of file from stat or lstat call - blank if posix_getpwuid is not available
        "groupName" : "foobar"       // (String) the posix group name for that gid - blank if posix_getgwgid is not available
    },
}
```

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -70,7 +70,8 @@ class elFinder {
 		'info'      => array('targets' => true),
 		'dim'       => array('target' => true),
 		'resize'    => array('target' => true, 'width' => true, 'height' => true, 'mode' => false, 'x' => false, 'y' => false, 'degree' => false),
-		'netmount'  => array('protocol' => true, 'host' => true, 'path' => false, 'port' => false, 'user' => true, 'pass' => true, 'alias' => false, 'options' => false)
+		'netmount'  => array('protocol' => true, 'host' => true, 'path' => false, 'port' => false, 'user' => true, 'pass' => true, 'alias' => false, 'options' => false),
+		'chmod'     => array('target' => true, 'mode' => true)
 	);
 	
 	/**
@@ -277,7 +278,6 @@ class elFinder {
 		$cmds = $cmd == '*'
 			? array_keys($this->commands)
 			: array_map('trim', explode(' ', $cmd));
-		
 		foreach ($cmds as $cmd) {
 			if ($cmd) {
 				if (!isset($this->listeners[$cmd])) {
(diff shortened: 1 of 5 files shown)


nao-pon:
@captin411 Thanks for your PR. I'll marge into 2.1 branch. But stat will be more simply like this.
```json
{
"isowner" : 1,
"perm" : "0755 or 755 or rwxr-xr-x",
"owner" : "name or UID",
"group" : "group or GID"
}
```

nao-pon:
Now chmod supported on 2.1 branch. ( diff: https://github.com/Studio-42/elFinder/compare/9b116e40e7f18c66afcf1abcb8fb15ab5137679c...60c784f4ab37b472d4446b3d644908a627be78a9 ) Thanks again !

PR #1030, opened 2015-06-11
change persian month names to english

tuytoosh:
in persian we have solar dates with persian month names...
for example today is 11,june,2015 equals 21,khordad,1394 in iran
elfinder is not supporting solar date and for example in elfinder today date is 11,khordad,2015 and it is not true

it is better to  change month names to English 
for example today is "11, زوئن(june) ,2015" and it is true


nao-pon:
@tuytoosh Thanks!

PR #1050, opened 2015-06-25
Use json_encode instead of serialize

barryvdh:
Using `serialize()` causes errors for Closures, see https://github.com/barryvdh/elfinder-builds/commit/e8357274137d60a6205d31df4840c8a1713b60c3#commitcomment-11854093 and https://github.com/barryvdh/elfinder-flysystem-driver/issues/10

> php/elFinder.class.php line 248
"Serialization of 'Closure' is not allowed"

json_encode should also be faster anyways. http://stackoverflow.com/questions/2254220/php-best-way-to-md5-multi-dimensional-array


nao-pon:
@barryvdh Thanks!

> JSON_ENCODE is consistently over 250% (2.5x) faster (often over 300%)

Wow! :+1:

Issue #1057, opened 2015-07-20
custom command and UI changes 


rut2prajapati:
I am working with elfinder for internal communication tool. I want to add functinalty in I can write description for each directory. For that I need to create commands by which I can store description text in database for perticular directory and then whenever that directory opens discription should be there. 

Is there any simple documant available where steps are defined about : How to create custom commands and change UI accordingly ?

lokothodida:
The [Adding file description to Properties dialog](https://github.com/Studio-42/elFinder/wiki/Adding-file-description-to-Properties-dialog) wiki page describes what you are trying to do. It's not an elegant solution however, as it involves editing the elFinder Javascript source code.

lokothodida:
There is now a [2.1 draft](https://github.com/Studio-42/elFinder/wiki/Adding-file-description-to-Properties-dialog-2.1) of this functionality which lets you achieve this in [elFinder 2.1](http://nao-pon.github.io/elFinder-nightly/latests/elfinder-2.1.zip) without editing the Javascript source code (the PHP is identical to before).

PR #1058, opened 2015-07-23
Update info.js

lokothodida:
[Adding extra fields to the properties dialog](https://github.com/Studio-42/elFinder/wiki/Adding-file-description-to-Properties-dialog) #1057 currently involves editing the source code in `js/commands/info.js`. This means any changes that a user makes here are incredibly brittle, as they need to be added and rebuilt into any new elFinder updates.

This change will allow users to add extra properties from the elFinder client configuration on the `commandsOptions` object:

```javascript
// In the configuration object:
commandsOptions : {
  info : {
    // Custom fields
    custom : {
      // Description field
      desc : {
        // Label
        label : 'Description',

        // Template
        tpl : '<div id="elfinder-fm-file-desc"></div>',

        // Restricts to mimetypes
        mimes : ['text/html'],

        // Request that asks for the description and sets the field
        action : function(file, filemanager, dialog) {
          filemanager.request({
            data : { cmd : 'desc', target: file.hash },
            preventDefault: true,
          })
          .done(function(data) {
            var $desc = dialog.find('#elfinder-fm-file-desc');
            $desc.html(data.desc);
          });
        },
      }
    }
  }
},
```

This protects users who want to add custom fields from digging around in the Javascript source code to make those changes.

### DIFF
--- a/js/commands/info.js
+++ b/js/commands/info.js
@@ -151,6 +151,18 @@ elFinder.prototype.commands.info = function() {
 			content.push(row.replace(l, msg.modify).replace(v, fm.formatDate(file)));
 			content.push(row.replace(l, msg.perms).replace(v, fm.formatPermissions(file)));
 			content.push(row.replace(l, msg.locked).replace(v, file.locked ? msg.yes : msg.no));
+
+			// Add custom info fields
+			if (o.custom) $.each(o.custom, function(name, details) {
+				if (!details.mimes || $.inArray(file.mime, details.mimes) >= 0) {
+					// Add to the content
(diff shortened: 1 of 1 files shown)


nao-pon:
@lokothodida :+1: I'll merge to 2.1 branch. Thanks!

nao-pon:
It is merged to 2.1 branch. Thanks!

-  82e34f75eaebe95c1450537c44bf6ad41df6e195 Merge remote-tracking branch 'lokothodida/patch-1' into 2.1
- e879dcd5a9ee5b20de1b7e4d0c86d129ef654dfe  [cmd:info] correction of custom column (#1058)

PR #1059, opened 2015-07-23
Import external js in a scheme independent way

chmduquesne:
This patches fixes browser warning/issues when serving from ssl (and keeps it working in http too)


megadr01d:
Notifying @nao-pon. I think this is a great addition to the code and prevents some users posting that elFinder can't find the backend.

nao-pon:
@chmduquesne , @megadr01d Thanks!

PR #1061, opened 2015-07-28
Update tree.js

lokothodida:
There is currently no way to add custom, directory-dependent CSS classes to the navbar icons. This change will allow this to be done on the elFinder client configuration object:

```javascript
// In the configuration object:
uiOptions : {
  tree : {
    cssClass: function(directory) {
      // This adds the directory's name (lowercase) as a CSS class
      return directory.name.toLowerCase();
    }
  }
},
// ...
```

Thus a client can style individual navbar directories based on the directory data (e.g. a 'Pictures' icon for a pictures folder, a 'Videos' icon for a videos folder, etc).

This also means that if the server (volume driver) provides custom data to the directory, that too can be used for styling: e.g. if the server gives the `directory` the property `foo`, a style can be added that is dependent on the value of `directory.foo`.

### DIFF
--- a/js/ui/tree.js
+++ b/js/ui/tree.js
@@ -180,7 +180,7 @@ $.fn.elfindertree = function(fm, opts) {
 			 */
 			replace = {
 				id          : function(dir) { return fm.navHash2Id(dir.hash) },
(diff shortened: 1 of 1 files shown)


nao-pon:
@lokothodida :+1: I think option name `getClass` is more better. Could you please change the name?

lokothodida:
:+1: Changed to `getClass`.

nao-pon:
Thanks! :smile:

PR #1069, opened 2015-08-07
Clean up source html file and update readme

lokothodida:
There are some major confusions for people starting out using elFinder:

* The repo's `elfinder.html` file doesn't work (it points to build files when it is for the source)
* `elfinder.src.html` is filled with tests that the average user doesn't want or need (and is badly marked up)
* One has to search the [issues](issues) to figure out which version of elFinder is best suited to them
* There aren't any real instructions for the initial installation of elFinder.

This is why so many people are stuck on version 2.0 (since it just works out of the box).

Changes:

* Copied the contents of `elfinder.src.html` to `elfinder.playground.html` and cleaned up the src file to only include relevant configuration
* Updated the readme to have pertinent information to installing elFinder

This should help newcomers with getting started. If these changes are implemented, they should be copied across to the 2.1 branch also.

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,6 +14,18 @@ elFinder is an open-source file manager for web, written in JavaScript using
 jQuery UI. Creation is inspired by simplicity and convenience of Finder program
 used in Mac OS X operating system.
 
+Contents
+--------
+* [Features](#features)
+* [Requirements](#requirements)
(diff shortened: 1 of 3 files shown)


nao-pon:
@lokothodida Nice work! and great! :+1: Thank you so much :-D

PR #1070, opened 2015-08-10
Change connector recommendation to minimal

lokothodida:
After some quick testing, its easiest for people to get started with the minimal connector even for the source version of elFinder, so I've changed that in the installation guide in the readme.

### DIFF
--- a/README.md
+++ b/README.md
@@ -79,7 +79,7 @@ Installation
       $ git clone https://github.com/Studio-42/elFinder.git
       ```
 
- 2. Rename `/php/connector.php-dist` to `/php/connector.php`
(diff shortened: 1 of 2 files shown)


Rillke:
You may want to [squash your commits](https://github.com/ginatrapani/todo.txt-android/wiki/Squash-All-Commits-Related-to-a-Single-Issue-into-a-Single-Commit) in this pull request into a single commit:
```
git checkout 2.1
git rebase -i HEAD~2
```
- s/pick/squash on "Update README.md"
- save and close
```
git push -f origin 2.1
```

Also, perhaps explain *why* the minimal example should be used by default instead.

nao-pon:
Thanks!

PR #1083, opened 2015-09-10
Make elFinderConnector::input_filter() protected

Rillke:
Given that almost all methods, `elFinderConnector` provides are already protected (or public), thus is meant to be sub-classed and `input_filter()` is useful in derived classes, too, make it protected instead of private.


nao-pon:
@Rillke That's right. Thanks! :+1:

PR #1100, opened 2015-10-07
Fix small semantic typo

MazeChaZer:


nao-pon:
@MazeChaZer Thanks! :+1:

PR #1125, opened 2015-11-10
Fix file permissions for MySQL LOAD_FILE command

MazeChaZer:
Without this `chmod()` call MySQL wasn't able to read the file.


nao-pon:
`$tmpfile` file made by `tempnam()` with permission `0600`. I propose a change to `chmod($tmpfile, 0644);`. So thank you to fix it if there is no problem.

MazeChaZer:
Yeah sure no problem, I amended the commit.

nao-pon:
@MazeChaZer Thanks! :+1:

PR #1130, opened 2015-11-13
Translated 2.1 keywords

tabacitu:

PR #1111, opened 2015-10-21
added Romanian translation

tabacitu:

PR #1132, opened 2015-11-13
Translated new 2.1 keywords in Romanian

tabacitu:


nao-pon:
@tabacitu Thanks a lot !

ref. #1111 , #1130

PR #1133, opened 2015-11-16
Additions to german translation

Timo-Linde:


nao-pon:
@Timo-Linde Thanks! :+1:

PR #1140, opened 2015-11-18
Plugin Normalizer option: make filename lowercase

zzplab:
I prefer making de filename lowercase always. I propose a pull request to have it build in the master in that way, that the user can decide tot turn the option on or off.
I put default to on, but default can be off to off course.

### DIFF
--- a/php/plugins/Normalizer/plugin.php
+++ b/php/plugins/Normalizer/plugin.php
@@ -24,7 +24,8 @@
  *			'Normalizer' => array(
  *				'enable' => true,
  *				'nfc'    => true,
- *				'nfkc'   => true
+ *				'nfkc'   => true,
+ *				'lowercase'   => false
  *			)
  *		),
  *		// each volume configure (optional)
@@ -37,7 +38,8 @@
  *					'Normalizer' => array(
  *						'enable' => true,
  *						'nfc'    => true,
- *						'nfkc'   => true
+ *						'nfkc'   => true,
+ * 						'lowercase'   => false
  *					)
  *				)
  *			)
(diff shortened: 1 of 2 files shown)


nao-pon:
@zzplab :+1: 

I think towards the Normalizer is more suitable to this function. Can you retrofit to Normalizer? In addition, the default value `false` I think is good.

Thanks! :smiley:

zzplab:
I 'm fine with Normalizer as the plugin for this function. In Sanitizer it was slightly easier to implement.
Reverted the Sanitizer option. Propose the changes in Normalizer.
Default value: false.
Note:
The $path should never be made lowercase I think, just the filename. Solved that, but let me know if you see a better way to prevent that.
Easier is to do the $name first and then the $path, but that could lead to issues if in the future if more code is added tot the plugin. In my proposal it is safer/foolproof. :-)
Thanks for the fast response.

nao-pon:
At first, You should be make new branch for pull request. Please do next steps on your local git repository.

- Close this Pull Request.

(# `patch-1` is any branch name. Please be your favorite.)
- git checkout `74cb43ae835d77005f5e354d5fcbc39ac0786c67`
- git checkout -b patch-1
- git cherry-pick `3ac4803693b04a2e6f559223ef4c03ef891055bc`
- git push origin patch-1

- And make pull request `zzplab/elFinder:patch-1` to `Studio-42/elFinder:2.x`.

If you want to reset your "2.x branch", you can in the next steps.

- git checkout 2.x
- git reset --hard `74cb43ae835d77005f5e354d5fcbc39ac0786c67`
- git push -f origin 2.x

Because it may keep track of the main stream it is difficult, it is better you have enforced the "one pull request / 1 branch".

PR #1142, opened 2015-11-22
Added disableInfoLink option. If set will not show directory/file link in Get Info popup

moshefi:


nao-pon:
@moshefi Please make new branch for pull request that has commits for new feature only. And Please commit without changing tabs and spaces and newline characters. Thanks!

Issue #1146, opened 2015-11-27
Page freeze when uploading files on elFinder 2.1.x


fauri13:
When I try to upload files the whole page freezes and doesn't respond. Version 2.0.4 works well, it's an issue with v 2.1.x (any version). 

It's something with the configuration of the server but don't know what could be...

I use last apache build and last php5 build.

Anyone else with the same issue?

nao-pon:
@fauri13 Try `error_reporting(-1);` in your connector.php for debugging. And check server response.

fauri13:
There are no errors at all in the connector. The post request never reaches the server. I have iptables configured but don't see either that the connection is blocked.

It's like the javascript enters an infinite loop... It starts to use huge amounts of CPU time and memory.

fauri13:
Found the error!

The connector sends to the client the value of upload_max_filesize in the php.ini. If it is set to 0, the uplMaxSize variable is set to "0" (as string, not numeric) and when getting the BYTES_PER_CHUNK, the or between "0" and a number, it returns "0". So 0 minus 8190 is negative and thus it enters an infinite loop.

Pushed the fix if you want to check it.

nao-pon:
@fauri13 Thanks! :+1:

PR #1148, opened 2015-11-28
Fix for issue #1146

fauri13:
Fix for issue #1146.
The connector sends to the client the value of upload_max_filesize in the php.ini. If it is set to 0, the uplMaxSize variable is set to "0" (as string, not numeric) and when getting the BYTES_PER_CHUNK, the or between "0" and a number, it returns "0". So 0 minus 8190 is negative and thus it enters an infinite loop.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -1029,7 +1029,7 @@ window.elFinder = function(node, opts) {
 	 * @return Integer
 	 */
 	this.returnBytes = function(val) {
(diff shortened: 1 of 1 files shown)


fauri13:
Changed the returnBytes funct so it always returns an integer (it was returning an string when val = "0").

I think it's a better solution than the first I made.

nao-pon:
Thank you very much! I'm going later to a little more maintenance. :+1:

PR #1158, opened 2015-12-04
Double @unlink() on temp file fix

MazeChaZer:
All temp files are already unlinked by the shutdown handler which is being registered in getTempFile().
For some reason, I got an error when the second unlink() in the handler tried to delete a file that was already deleted by the first unlink() call (although both unlink calls are prefixed with an '@').

### DIFF
--- a/php/elFinderVolumeDropbox.class.php
+++ b/php/elFinderVolumeDropbox.class.php
@@ -1146,9 +1146,6 @@ class elFinderVolumeDropbox extends elFinderVolumeDriver {
 	 **/
 	protected function _fclose($fp, $path='') {
 		@fclose($fp);
-		if ($path) {
-			@unlink($this->getTempFile($path));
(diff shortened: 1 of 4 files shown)


nao-pon:
@MazeChaZer Unlink in shutdown function is a ancillary. Therefore,  do file check before unlink there. - 0091197c950d6b702e6e604225908754149bfa27

What about this?

MazeChaZer:
Your fix looks good and should fix my issue, thanks.
Although in the class `elFinderVolumeLocalFileSystem` the `unlink()` calls seems to be missing in the `_fclose()` method.

nao-pon:
LocalFileSystem driver does not create a temporary file in `_fopen()`, so I think there is no need. ;-)

MazeChaZer:
Oh yeah I see, that makes sense :)

PR #1218, opened 2016-02-14
Typo fix

flack:
According to https://github.com/Studio-42/elFinder/blob/master/js/elFinder.options.js#L481, the variable needs to be called `sortRules`


nao-pon:
@flack Thanks! :+1:

PR #1238, opened 2016-03-02
Small l10n tweaks

flack:
Since German doesn't use am/pm, it's better to show hours in 24h format (desktop file managers do that, too). Also fixes a wrong capitalization


nao-pon:
@flack Thanks!

PR #1249, opened 2016-03-09
Fix undefined variable

barryvdh:
Not sure if this is correct. Either way, $file is sometimes not set. So either check `isset($file)` before doing that, or use a different variable. Otherwise this will result in an Exception about undefined variable.
See https://github.com/barryvdh/laravel-elfinder/issues/130#issuecomment-193628280

### DIFF
--- a/php/elFinderVolumeDriver.class.php
+++ b/php/elFinderVolumeDriver.class.php
@@ -3375,7 +3375,7 @@ abstract class elFinderVolumeDriver {
 			if (elFinder::$defaultMimefile) {
 				$file = elFinder::$defaultMimefile;
 			}
-			if (! is_readable($file)) {
+			if (! is_readable($path)) {
 				$file = dirname(__FILE__).DIRECTORY_SEPARATOR.'mime.types';
 			}
 			if (is_readable($file)) {


nao-pon:
oops! @barryvdh Thank you veeeery much!

barryvdh:
Can you tag a new release sometime soon? Now latest release is broken :(

nao-pon:
@barryvdh Yes, OK. I’ll tagging in a soon. Maybe today or tomorrow.

barryvdh:
Sorry I don't think this is fully correct either. We're still not sure if $file is set in the next line.
Not sure if it's $file or $path, but we should check if they're set before using them, or initialize at null first.

nao-pon:
@barryvdh [This code](https://github.com/Studio-42/elFinder/blob/b5be56bed94bb7bcccaa0fd2d134329325963ea1/php/elFinderVolumeDriver.class.php#L3375) is
```php
	static protected function mimetypeInternalDetect($path = '') {
		// load default MIME table file "mime.types"
		if (!elFinderVolumeDriver::$mimetypesLoaded) {
			elFinderVolumeDriver::$mimetypesLoaded = true;
			$file = elFinder::$defaultMimefile;
			if ($file === '' || ! is_readable($file)) {
				$file = dirname(__FILE__).DIRECTORY_SEPARATOR.'mime.types';
			}
			if (is_readable($file)) {
				$mimecf = file($file);
				foreach ($mimecf as $line_num => $line) {
					if (!preg_match('/^\s*#/', $line)) {
						$mime = preg_split('/\s+/', $line, -1, PREG_SPLIT_NO_EMPTY);
						for ($i = 1, $size = count($mime); $i < $size ; $i++) {
							if (!isset(elFinderVolumeDriver::$mimetypes[$mime[$i]])) {
								elFinderVolumeDriver::$mimetypes[$mime[$i]] = $mime[0];
							}
						}
					}
				}
			}
		}
		$ext = '';
		if ($path) {
			$pinfo = pathinfo($path); 
			$ext   = isset($pinfo['extension']) ? strtolower($pinfo['extension']) : '';
		}
		return ($ext && isset(elFinderVolumeDriver::$mimetypes[$ext])) ? elFinderVolumeDriver::$mimetypes[$ext] : 'unknown';
	}
```

$file initialized by `$file = elFinder::$defaultMimefile;`.

barryvdh:
Oh you already fixed it :) Sorry didn't see that.

nao-pon:
@barryvdh Thanks! But [chrisjohnstone](https://github.com/barryvdh/laravel-elfinder/issues/133#issuecomment-195291789) say 2.1.9 has a problem about "Array to string conversion error". I don't  see this problem. Do you know anything?

bertalanimre:
Hello,

All I can say I also get this error whenever I want to upload an image. First it was Undefined variable: File, after a composer update it changed to: Array to string conversion

Any idea?

nao-pon:
The cause of the "Array to string conversion" error is because the old JavaScript of elFinder.

Issue #1285, opened 2016-03-24
Error moving file from one GCS bucket (in sub-folder) to another bucket (root folder)


paulcanning:
NB - I have posted this here as the file giving me the error is from the elFinder base library.

I get the following error when trying to move a file from "my-bucket/sub-doler/file.txt" to "my-other-bucket/file.txt"

Fatal error: Nesting level too deep - recursive dependency? in C:\wamp\www\my-app\application\libraries\elfinder\elFinderVolumeDriver.class.php on line 1887

Update - I get the same error trying when no sub-folder is involved.

paulcanning:
I fixed the issue, I'll do a pull request

nao-pon:
@paulcanning Please send PR to master.

PR #1287, opened 2016-03-24
Fix for issue #1285

paulcanning:


nao-pon:
@paulcanning Thanks! :+1:

PR #1289, opened 2016-03-26
Uyghur translation

tugh:
Hi there, added Uyghur translation. hope you can merge it.

Thanks.


nao-pon:
@tugh Thanks! Is it needed "_CN"? I'm sorry if it is rude. Because I don't know well.

tugh:
@nao-pon
It's ok man. The iso standards is like that, so yeah it should go with "_CN".

nao-pon:
@tugh Thanks! :+1:

PR #1290, opened 2016-03-26
Updated french translations

gabsource:
Translations for new messages


nao-pon:
@gabsource Thanks! :+1:

PR #1333, opened 2016-04-19
use npm for installing build dependencies

troex:
Now you can simply use `npm install` which will install nodejs packages required to build elfinder, also by running `npm run-script build` will build elfinder into ./build dir.

Once `npm install` is run it will install npm packages into `./node_modules` by default. You can execute jake by running `./node_modules/.bin/jake`.

This is the default way to go with npm.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,0 +1,33 @@
+{
+  "name": "elFinder",
+  "version": "2.1.11",
+  "description": "File manager for web",
+  "main": "Jakefile.js",
+  "scripts": {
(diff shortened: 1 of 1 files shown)


nao-pon:
@troex It should be described version information always?

troex:
I'm not sure, since we are not pushing our project to npmjs.org maybe we can omit version...

troex:
removing `version` doesn't hurt npm, so I removed it

nao-pon:
@troex Thank you for doing this. :+1:

PR #1348, opened 2016-04-28
pt_BR Translation fixes

fhb12345:
Fixed some misspelled phrases and words.


nao-pon:
@fhb12345 Thanks! :+1:

PR #1349, opened 2016-04-28
Mounting Google Drive with elFinder

rajasharma27:
Mounting Google Drive in elFinder with google-api-php-client-2.0.0-RC7 or above

### DIFF
--- a/php/elFinderVolumeGoogleDrive.class.php
+++ b/php/elFinderVolumeGoogleDrive.class.php
@@ -1,0 +1,1602 @@
+<?php
+elFinder::$netDrivers['googledrive'] = 'GoogleDrive';
+
+/**
+ * Simple elFinder driver for GoogleDrive
+ * google-api-php-client-2.x or above
+ * requird PHP 5.6.20 or above
+ *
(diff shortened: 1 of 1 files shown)


nao-pon:
@rajasharma27 From where the First noticed...
Please omit the unnecessary things about the options. Also, please let versatility without defining a value. For example, such as callback or path.

rajasharma27:
sir, I want user give only connector url in the redirecturi,
 if (isset($_GET['code'])) {
	$callback = new elFinderVolumeGoogleDrive();
	$callback->netmountPrepare(array());  //prepare callback url
    exit();
  }
when remove this from connector, drive class not read the value from define variable from connector class, because these variable define after this
include_once dirname(__FILE__).DIRECTORY_SEPARATOR.'elFinderVolumeGoogleDrive.class.php';

suggest me

rajasharma27:
according to your code
// // AND reuire regist redirect url to "YOUR_CONNECTOR_URL?cmd=netmount&protocol=googledrive&host=1"

working without callback or path

rajasharma27:
thanks

PR #1350, opened 2016-04-28
MSSql table structure for mounting in elFinder

rajasharma27:
MSSql table structure for mounting in elFinder

### DIFF
--- a/php/MsSQLStorage.sql
+++ b/php/MsSQLStorage.sql
@@ -1,0 +1,64 @@
+BEGIN
+IF EXISTS (SELECT 1 
+           FROM INFORMATION_SCHEMA.TABLES 
+           WHERE TABLE_TYPE='BASE TABLE' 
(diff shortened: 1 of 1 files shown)


nao-pon:
@rajasharma27 I do not know much about MsSql. Is this thing generic? (e.g. "DF__elfinder_f__mime__7F60ED59")

rajasharma27:
actually 7F60ED59 this type of value I found somewhere, not clear use in constraints name

nao-pon:
@rajasharma27 I don't know well about MsSql. Please make clearly. And, SQL query "ALTER TABLE" can be included in "CREATE TABLE"?

nao-pon:
There is a regrettable, but we drop the MsSQL driver. Because it can not be maintained to continue because I am not familiar with MsSQL.

So you can listed in the README as a third-party volume driver, please PR to README.md.

PR #1352, opened 2016-04-28
connector configuration for GoogleDrive

rajasharma27:
connector configuration for GoogleDrive


nao-pon:
OAuth callback, please be treated with GoogleDrive volume driver. And, please integrated into the #1349.

PR #1429, opened 2016-05-27
Time formatting showing AM for h = 12 where PM should be shown

japerr:
Changing to h >= 12 shows PM correctly. Change made for both 'a' and 'A' cases.


nao-pon:
@japerr Yes, that's right! Thanks! :+1:

PR #1442, opened 2016-06-02
Updated Slovak translation

RobiNN1:


nao-pon:
@RobiNN1 Thanks! :+1:

PR #1519, opened 2016-07-11
Updated italian translation

nicorac:
Here you ahe the updated italian translation for elFinder.
Hope you'll merge it.

Thanks for your great work.
Claudio


nao-pon:
@nicorac Thanks! :+1:

PR #1527, opened 2016-07-13
elFinderVolumeDriver::realpath return path if file not exist

kukko:
Added a parameter to the realpath method. If it is setted to true, the method return the path if file not existing. It is required because when elFinder tried to write log about rename folder/file, at the moment of logging, the file which is renamed is not existing.


nao-pon:
@kukko I pushed @f638d9b748657d3ed01b7a4a009bf6b5f0b9de42. Now you can use `$volume->getPath($hash)` in bind function.

e.g.
```php
function logger($cmd, $result, $args, $elfinder, $volume) {
	if (! empty($result['removed'])) {
		$removed0 = $result['removed'][0];
		$path0 = $volume->getPath($removed0);
	}
}
```

Issue #1572, opened 2016-08-05
Unable to drag drop a folder named "0"


marc-chappuis:
When uploading a folder named "0", there is an alert with the message:

Unable to create folder "".
Invalid file name.

The alert is also displayed twice

![elfinder-issue](https://cloud.githubusercontent.com/assets/20840815/17431252/61b9e7fc-5af9-11e6-80d4-1e7c81dbd2fc.png)

To reproduce the bug, create a folder on your computer named "0", add at least one file inside and
drag drop the folder to elfinder.

If you create a folder named "0" directly on elfinder it works.

If you drag drop a folder named "0" from your computer on an existing one on elfinder, it works after asking you if you want to replace it.

Was using chrome 51.0.2704.103 on windows 10

nao-pon:
@marc-chappuis This is a bug of elFinder.class.php. I'll fix it.

PR #1574, opened 2016-08-05
solve #1572

marc-chappuis:
json_encode(0) was evaluated as false


nao-pon:
@marc-chappuis Thanks! :+1:

PR #1587, opened 2016-08-08
Support the attribute "ts"

dup2:
Up to now, the modification time of an object was only transported via the "date" attribute and this was also used for sorting.

The "date" attribute was a string, formatted and localised - i.e. "23.05.2016 18:34". For sorting, this does mostly not work as the sort function compares strings and this leads to a wrong order.

Instead, elfinder pefers the "ts" attribute which can be used for sorting easily (numerical value) and is then formatted client-side with JS to be displayed.

This patch allows for an old API server to send the "ts" attribute and thus fix the sorting issue.


nao-pon:
@dup2 I have a plan that convert date string to `ts`. I'll commit and push soon.

Thanks! :+1:

PR #1616, opened 2016-08-24
2.1

dvpweb:


nao-pon:
@dvpweb It is not possible to accept the customized for a particular volume driver. Please correspond create your own class inherits the elFinder class.

Thanks for the proposal.

PR #1631, opened 2016-08-29
2.1 src, pull request from jonny at tiki.org

jonnybradley:
Hi Studio-42

I've been maintaining a fork of elFinder for a few years now for use in Tiki (tiki.org) and have made a few fixes i would like to keep, is this pull request enough for you to go on?

Sorry, still very much a git-n00b, obviously the change to the version string shouldn't be here and sorry about all the whitespace changes, i have cleaned them out 3 times now and they seem to keep coming back (or i can't commit them as there's no changes to commit) - grrr!

Anyway, hope this helps and thanks for all your great work everybody! :)

jb

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,4 @@ elFinderVolumeSVN.class.php
 connector-svn.php
 node_modules
 connector.php
+.idea
--- a/css/dialog.css
+++ b/css/dialog.css
@@ -10,6 +10,11 @@
 	right:auto;
 }
 
+/* position getting overwritten as relative by jquery-ui 1.10 */
+.std42-dialog.elfinder-dialog {
+	position:absolute;
+}
+
 /* titlebar */
 .std42-dialog .ui-dialog-titlebar {
 	border-left:0 solid transparent;
--- a/css/dialog.css
+++ b/css/dialog.css
@@ -99,11 +99,11 @@
 
 /*********************** ERROR DIALOG **************************/
 
-.elfinder-dialog-error .ui-dialog-content,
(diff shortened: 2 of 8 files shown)


nao-pon:
@jonnybradley Thanks for your request. However, These modifications looks to those of your project-specific.

Of course, the requests required for stand-alone elFinder are welcome. Please make pull request to master branch from each branch of the problem.

jonnybradley:
Hi @nao-pon - thanks for looking at them so quickly.
I'm not sure how they are too project (tiki) specific, maybe my commit messages aren't any good :) Can i maybe explain them a little more?

- Adding the startroot option seems pretty generic for any implementation using more than one volume and wanting to use any but the first one by default, or is Tiki the only project doing this?
- Refresh current dir after extracting: Does this get done automatically somehow in other projects? For us it still shows the directory contents before the files were extracted, i'll look at this again.

I guess the other CSS ones are probably not generic enough, and as i said i'm new to git pull requests, so i guess i need to make a clean branch and only pick the relevant commits, mainly the startroot one i guess.

Thanks again, more soon... :)

nao-pon:
>
- Adding the startroot option seems pretty generic for any implementation using more than one volume and wanting to use any but the first one by default, or is Tiki the only project doing this?

Please set the value to startpath only volume to which you apply the startPath.

>
- Refresh current dir after extracting: Does this get done automatically somehow in other projects? For us it still shows the directory contents before the files were extracted, i'll look at this again.

There is no problem in the standard drivers such as LocalVolumeSystem. I guess that there is a cause in customized drivers.

PR #1683, opened 2016-09-30
Avoid re-enable disabled buttons

dwickstrom:
...on searchend event

Not knowing what the expected behaviour is suppose to be, I perceived this as a bug. Here's how to reproduce:

1. Click on a folder where some buttons are disabled due to restrictions
2. Those buttons are now disabled
3. Click on the active folder (again)
4. Those buttons are now enabled (by removed class 'ui-state-disabled')


nao-pon:
@dwickstrom Thanks!

PR #1711, opened 2016-10-21
exclude php folder from bower

Pchol:


nao-pon:
#1731 is my mistake. That was canceled.

PR #1731, opened 2016-11-03
Revert "exclude php folder from bower"

nao-pon:
Reverts Studio-42/elFinder#1711


nao-pon:
This is my commit mistakes. This was canceled.

PR #1771, opened 2016-11-25
Removed an unused variable

taylankasap:


nao-pon:
@taylankasap Thanks! 👍

PR #1773, opened 2016-11-25
Fix for text file corruption after edit if encoding is different from UTF-8

vvzh:
I have elFinder 2.1 running on Windows server with Cyrillic (CP1251) codepage.

In order to get proper filesysem encoding of files and folders I create via web UI I have to specify 'encoding' as 'CP1251' in connector configuration. This also makes elFinder assume text files are stored in CP1251 too, that is fine for me.

Then, if I edit such a text file via web UI, it is loaded as CP1251-encoded but saved as UTF-8-encoded. elFinder cannot properly display this file the second time I open it: text becomes corrupted because UTF-8 data is interpreted as being CP1251. Saving the file writes down this corrupted content to disk.

This pull request fixes the problem.

### DIFF
--- a/php/elFinderVolumeDriver.class.php
+++ b/php/elFinderVolumeDriver.class.php
@@ -2352,7 +2352,7 @@ abstract class elFinderVolumeDriver {
 		}
 		
 		$this->clearcache();
-		return $this->convEncOut($this->_filePutContents($this->convEncIn($path), $content)) ? $this->stat($path) : false;
+		return $this->convEncOut($this->_filePutContents($this->convEncIn($path), $this->convEncIn($content))) ? $this->stat($path) : false;
 	}
 
 	/**


nao-pon:
@vvzh Since server encoding and content encoding do not necessarily match, we can not accept this request.

However, there is a way to use the bind option as a solution.

```php
function putConvEncord($cmd, &$args, $elfinder, $volume) {
    if (! empty($args['content'])) {
        $args['content'] = $volume->convEncIn($args['content']);
    }
}
$opts = array(
    'bind' = array(
        'put.pre' = array('putConvEncord'),
    ),
    'roots' => array(...),
);
```

I appreciate your contribution. 👍

vvzh:
Well, I understand the reason. Thanks for bind option example, it is much better than hardcoding file encoding to be equal with server one.

I made an assumption about content encoding matching server encoding because getContents function already encodes file contents using server encoding:
https://github.com/Studio-42/elFinder/blob/master/php/elFinderVolumeDriver.class.php#L2311
Maybe it's incorrect too and needs a fix?

nao-pon:
> Maybe it's incorrect too and needs a fix?

Yes, That is a bug. I'll fix it. Thanks! 👍

PR #1801, opened 2016-12-10
Improve zh-tw localization

danny0838:
Fix an issue that zh-TW be detected as zh_CN.

Add new localization entries as well as fix old ones.


nao-pon:
@danny0838 Thanks!  谢谢！ 非常感谢 👍

Issue #1800, opened 2016-12-10
PHP Warning after update to 2.1.19


flack:
I've recently updated from 2.1.16 to 2.1.19 and since then, I get this error sporadically in JSON responses:

```
Warning: session_start(): Cannot send session cookie - headers already sent by (output started at vendor/studio-42/elfinder/php/elFinderConnector.class.php:317) in vendor/studio-42/elfinder/php/elFinderSession.php on line 47
```

EDIT: The last working version for me was actually 2.1.16, I've corrected the description above. This only happens on a PHP 5.3 server currently, I can't reproduce unter PHP 5.6

nao-pon:
@flack I'm checking on PHP 5.3.29 server. But I can't reproduce of this problem yet.

I tried next steps.

+ Upload elFider 2.1.19 to PHP 5.3.29 server
+ Rename connector.minimal.php.dist to connector.minimal.php
+ Make new file
+ Upload JPEG image file
+ Dupulicate image file
+ Remove image file
+ Edit text file

That's no problem.

flack:
I'm on PHP 5.3.3 (Debian 6) and I'm using a custom adapter (extends `elFinderVolumeDriver`, source is available here: https://github.com/flack/openpsa/blob/master/lib/org/openpsa/documents/elFinderVolumeOpenpsa.php), so maybe this has something to do with it. But it worked fine since at least 2.1.8, until 2.1.16 (2.1.17 doesn't run at all on PHP 5.3, so I don't know if this would also exhibit the problem)

nao-pon:
It's in not clear but I'm guessing that registered `midcom::get()->finish();` in the shutdown function may have some influence.

flack:
Thanks for looking into the issue!

I've tried removing the `midcom::get()->finish();` line, and also completely removing the shutdown function, but the warning still appears

flack:
I've bisected the issue now, and it looks like it comes from this commit: 7b8053a3bac1b3602830a04ade5a1f0150b708c1

flack:
I've tried setting `noSessionCache` for the volume, but it didn't help. The problem goes away however when I remove this line: https://github.com/Studio-42/elFinder/blob/master/php/elFinderVolumeDriver.class.php#L675

nao-pon:
What about next?

elFinderVolumeDriver.class.php
```php
	public function __destruct() {
		if (! headers_sent()) {
			$this->session->set($this->id, $this->sessionCache);
		}
	}
[...]

PR #1803, opened 2016-12-11
use session_id check for existing session in php5.3

flack:
This fixes the warning in #1800

### DIFF
--- a/php/elFinderSession.php
+++ b/php/elFinderSession.php
@@ -42,7 +42,7 @@ class elFinderSession implements elFinderSessionInterface
 			if (session_status() !== PHP_SESSION_ACTIVE) {
 				session_start();
 			}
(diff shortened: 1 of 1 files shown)


nao-pon:
@flack Thanks for your contribute. But becouse session_id() return value is not '' when session closed,  I can't accept this PR. Also If session_start () is called while session exists, a Notice error will occur.

**Test code**
```php
<?php
error_reporting(-1);

function session_start_error ($errno , $errstr) {}

$res = array();

$res['nothing session'] = session_id();

session_start();

set_error_handler('session_start_error',  E_NOTICE);

session_start();

restore_error_handler();

$res['session started'] = session_id();

session_write_close();

$res['session closed'] = session_id();

var_dump($res);
```

PR #1805, opened 2016-12-12
Fixed italian translation of english "Cancel"

nicorac:
Italian translation "Cancellare" could generate confusion, replaced with "Eliminare" (Remove/Delete) and "Annullare" (Cancel/Abort)


nao-pon:
@nicorac Thanks! 👍

PR #1811, opened 2016-12-15
Middleware for handling events

melanholly:
I am developing native QT desktop app and it includes file system calls and some custom commands. While developing it I came accross a problem. I needed authentication so some files can be locked with password. Everything was going good as planned until I discovered that I can't limit the user ability to expand the tree. So I decided to implement this simple middleware functionality. 

It can be used to stop the function from execution. For example to input some kind of password or confirm something. 

Here is a quick example:

```javascript
var elfinderInitOptions = {
    url       : 'elfinder-connector',  // connector URL (REQUIRED)
    middleware: {
        tree: {
            clickArrow: [function (next, context, arrow, link, stree, dfrd) {
                if (!link.hasClass('elfinder-subtree-loaded')) {
                    console.log('We do not continue');
                    return true;
                }
                console.log('calling next function in the middleware chain');
                return next();
            }]
        }
    }
}
```
In order for this to work **YOU HAVE TO return `next()`**.


nao-pon:
@melanholly Can that be handled by extending `fm.request ()`?

e.g. ([online demo](http://jsbin.com/sovuri/edit?js,output))
```javascript
var fm = $('#elfinder').elfinder(opts).elfinder('instance');
var _fmRequest = fm.request;
fm.request = function(opts) {
	var cmd = (opts.data || opts).cmd;
	if (cmd === 'open' || cmd === 'tree') {
		if (! window.confirm('Are you OK?')) {
			fm.trigger(cmd + 'done');
			fm.error('It was canceled by you.');
			if (cmd === 'tree') {
				setTimeout(function() {
					// restore arrow state
					$('#' + fm.navHash2Id(opts.target))
						.removeClass('elfinder-subtree-loaded')
						.addClass(fm.res('class', 'navcollapse'));
				}, 10);
			}
			return $.Deferred().reject();
		}
	}
	return _fmRequest.call(fm, opts);
};
```

melanholly:
Your code seems like a good example but this way the `fm.request` function will grow in complexity every time you need a custom hook in for some `cmd`. If the middleware approach is used you can maintain your functions in separated _small code blocks_ dedicated to specific events. 

I have added the middleware approach to the `tree` component only because it was simpler to use it for an example. Basily I plan to add it to all the function blocks of elfinder if it is good enough for it.

nao-pon:
I think that it is sufficient in many cases to cancel the request to the back end. By triggering an event from the request function, you can do that with elFinder 's bind function.

In the last sample it looks like the following. ([online demo](http://jsbin.com/duzuloq/edit?js,output))
```javascript
fm.bind('exec.open exec.tree', function(e) {
    var cmd = e.type.substr(5),
        opts = e.data.opts;
    if (! window.confirm('Are you OK?')) {
        fm.trigger(cmd + 'done');
        fm.error('It was canceled by you.');
        if (cmd === 'tree') {
            setTimeout(function() {
                // restore arrow state
                $('#' + fm.navHash2Id(opts.target))
                    .removeClass('elfinder-subtree-loaded')
                    .addClass(fm.res('class', 'navcollapse'));
            }, 10);
        }
        e.data.result = false;
    }
});
```
It is more complicated to set hook points for middleware in various parts of elFinder. Also, you have to think about what kind of parameters to pass each time.

How is the proposed event trigger from this request function? If that's OK, I'll add an event trigger point to the request function.

nao-pon:
And, of course, it also supports asynchronous processing. ([live demo](http://jsbin.com/nacoti/edit?js,output))

```javascript
fm.bind('exec.open exec.tree', function(e) {
    var dfrd = $.Deferred(),
        cmd = e.type.substr(5),
        opts = e.data.opts;
    if (! opts._1811confirm) {
        fm.confirm({
            title : '#1811 Test',
            text  : 'Are you continue?',
            accept: {
                label    : 'btnYes',
                callback : function() {
                    dfrd.resolve();
                }
            },
            cancel: {
                label    : 'btnNo',
                callback : function() {
                    dfrd.reject();
                    if (cmd === 'tree') {
                        setTimeout(function() {
                            // restore arrow state
                            $('#' + fm.navHash2Id(opts.target))
                                .removeClass('elfinder-subtree-loaded')
                                .addClass(fm.res('class', 'navcollapse'));
                        }, 10);
                    }
                }
            }
        });
        e.data.result = dfrd;
    }
});
```

melanholly:
Wow I have no Idea that there is a `bind` function. It is just what I need. I will close this pull request since this functionality already exists. 

Please excuse me for taking your time and not reading the documentation properly.

nao-pon:
@melanholly Thanks for your contribution. 👍

PR #1820, opened 2016-12-22
Added "sound" option

romaricdrigon:
Hello,

Thank you for this great library!

In our project, we don't want sounds at all.
In this PR, I propose to add a `sound` option (default `true` so BC is ok), you can use to disable those.
I don't know how to modify Wiki, I guess it is on your side. I may submit another PR, any feedback about the process is welcome :)


dup2:
+1

nao-pon:
@romaricdrigon Thanks for your contribution! 👍

PR #1821, opened 2016-12-22
[WIP] Added a "disabledInside" option

romaricdrigon:
Hello,

On an item returned by the connector, you can set some commands you want to disable.
This PR adds a similar `disabledInside` option, that allows to disable some commands when this item is your current working dir.

It does not overlap with `disabled` option at all, you can see as an option to disable commands when you click in the free space (like making a new folder...). Also that way BC is preserved.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -108,19 +108,20 @@ var elFinder = function(node, opts) {
 		 * @type Object
 		 **/
 		cwdOptions = {
-			path          : '',
-			url           : '',
-			tmbUrl        : '',
-			disabled      : [],
-			separator     : '/',
-			archives      : [],
-			extract       : [],
-			copyOverwrite : true,
+			path            : '',
+			url             : '',
+			tmbUrl          : '',
+			disabled        : [],
+			disabledInside  : [],
+			separator       : '/',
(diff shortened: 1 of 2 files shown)


romaricdrigon:
I added a "wip" flag, as for now there are 2 issues I need to fix:
 * it should be disabled in the toolbar too
 * the disable can not be reverted

romaricdrigon:
The issue about the non-revertable disable is fixed. About the toolbar, I think I need help - it looks like there are quite a few ways commands are checked for activation?
Inside `this.isCommandEnabled`, `cwdOptions.disabled` is always empty.

nao-pon:
@romaricdrigon I think that it is better to use the disabled option instead of increasing the new setting. To do that, I think that cwdOptions should be able to specify each option specific to each folder, and volOptions will need to correspond to each folder specification.

I will consider how to implement it. I will open a new issue when I found a solution.

nao-pon:
I opened new a issue #1826. so I'll close this when #1826 fixed.

romaricdrigon:
Thank you very much @nao-pon for reacting so quickly and working on this, I will check out your solution :)

PR #1829, opened 2016-12-29
Fixed typo in getDisabledCmds()

romaricdrigon:
Hi,

Thank you very much for working on the "disabled commands" option. I'm currently testing it.
There was a typo, that could pop in some case, here it is fixed.


nao-pon:
Oops! Oh my... Thanks so much @romaricdrigon !

Issue #1830, opened 2016-12-29
Allow to specify disabled commands by file


romaricdrigon:
Hello,

Since #1826 we can disable commands for a directory, which is great.
But then every file in this directory will have the specified commands disabled. IMO, this is too much. For better flexibility, files should also be able to have some `disabled`commands attached, and by default won't inherit those from parent. That will prevent side effects.

I'm working on this, but slowly, any help appreciated :)

romaricdrigon:
Done, I'm closing it!

Issue #1826, opened 2016-12-25
[js:core] allow command invalidation by each item(folder/file)


nao-pon:
c.f. #1821 

It is possible by the volume driver returning item stat data (Array) `folderstat.options.disabled`.

nao-pon:
**Example to disable `mkdir`, `paste` of folder name "cmdtest" of LocalFileSystem**

```php
class elFinderVolumeMyLocalFileSystem extends elFinderVolumeLocalFileSystem
{
    protected function updateCache($path, $stat) {
        if (! empty($stat) && is_array($stat) && $stat['mime'] === 'directory') {
            $name = $this->_basename($path);
            if ($name === 'cmdtest') {
                if (! isset($stat['options'])) {
                    $stat['options'] = array();
                }
                if (! isset($stat['options']['disabled'])) {
                    $stat['options']['disabled'] = $this->disabled;
                }
                array_push($stat['options']['disabled'], 'mkdir', 'paste');
            }
        }
        return parent::updateCache($path, $stat);
    }
    public function mkdir($dsthash, $name) {
        if ($dir = $this->dir($dsthash)) {
            if ($dir['name'] === 'cmdtest') {
                return $this->setError(elFinder::ERROR_PERM_DENIED);
            }
        }
        return parent::mkdir($dsthash, $name);
    }
    public function paste($volume, $src, $dst, $rmSrc = false, $hashes = array()) {
        if ($dir = $this->dir($dst)) {
            if ($dir['name'] === 'cmdtest') {
                return $this->setError(elFinder::ERROR_PERM_DENIED);
            }
        }
        return parent::paste($volume, $src, $dst, $rmSrc, $hashes);
    }
}
```

PR #1831, opened 2016-12-29
Each file/folder can have its own options

romaricdrigon:
This is a generalization of #1826, responding to #1830 : every file or folder can specify its own option. If none is specified, the same behavior that before is used (looking at volume for directories...).

For me, it was not making sense to be able to do something on directories and not files, since the API between those 2 is consistent right now. Plus this allows a better granularity to control which commands are available.


nao-pon:
@romaricdrigon Thanks!

PR #1857, opened 2017-01-18
Check if session exists in php < 5.4.0

devsseb:
If set_error_handler() function is already used before in custom code in connector.minimal.php

### DIFF
--- a/php/elFinderSession.php
+++ b/php/elFinderSession.php
@@ -45,9 +45,9 @@ class elFinderSession implements elFinderSessionInterface
 				session_start();
 			}
 		} else {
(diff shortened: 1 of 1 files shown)


nao-pon:
@devsseb elFinder may close the session early in order to improve performance during multi session and reopen it if necessary. But `session_id()` can not determine closed state. Therefore, I can not accept this PR.

**Test code**
```php
<?php
error_reporting(0);
header('Content-type: text/plain');

var_dump(session_id());

echo "session start\n";
session_start();
var_dump(session_id());

echo "session close\n";
session_write_close();
var_dump(session_id());
```

I appreciate your contribution.

devsseb:
Thanks you for your explanations

PR #1868, opened 2017-01-25
commands/download: fixed up timeouts for bad connection download

johnynsk:
We got issues from customers about download problem:
some ftps have an authorization lags, which are about one second.

When they are downloading file from slow ftp, they have got cancelled download task (frame was removed).

This fix keeps firefox timeouts. There are no any reason why we should remove frame after one second from file download.


nao-pon:
@johnynsk Thanks!

PR #1883, opened 2017-02-03
Update elfinder.zh_CN.js

changkejun:
I added some trans of CN.


nao-pon:
@changkejun Thanks! 👍

Issue #1895, opened 2017-02-14
hover css does not include 2nd line


Flamenco:
<img width="305" alt="hover issue" src="https://cloud.githubusercontent.com/assets/819940/22910266/cc7a7b96-f226-11e6-88c2-c388db85f58a.png">

This fixes it for me (changing height to min/max)
```css
/* file container */
.elfinder-cwd-view-icons .elfinder-cwd-file {
	width:120px;
	/*height:80px;*/
	min-height: 80px;
	max-height: 100px;
	padding-bottom:2px;
	cursor:default;
	border:none;
/*	overflow:hidden;*/
	position:relative;
}
```

And the results
<img width="242" alt="screen shot 2017-02-13 at 7 54 43 pm" src="https://cloud.githubusercontent.com/assets/819940/22910297/05ab7afa-f227-11e6-8d4d-6f817000281a.png">

<img width="333" alt="screen shot 2017-02-13 at 7 54 39 pm" src="https://cloud.githubusercontent.com/assets/819940/22910300/090a9a46-f227-11e6-81b3-376a13b09b9e.png">

PR #1896, opened 2017-02-14
Allow file icon height to be dynamic (for hovering)

Flamenco:
see https://github.com/Studio-42/elFinder/issues/1895


Flamenco:
Better solution may be to set height to 90px.  This solves the issue.  Dynamic height causes this:

screen shot 2017-02-13 at 8 14 04 pm

Flamenco:
Any thoughts on how to use min-height / max-height without the gap in the screenshot above?

nao-pon:
@Flamenco Thanks!

PR #2026, opened 2017-05-28
Update elfinder.id.js

ammarfaizi2:
Update elfinder.id.js


nao-pon:
@ammarfaizi2 Thanks! 👍

PR #2089, opened 2017-07-04
Add support for X-Sendfile and similar file-download accelerators

### LINKED ISSUE #475: Support Apache XSendFile module
It's like an enhancemet. When I start to use you elFinder with big files there appeared some problems with download of big files. I've started to research it and find for my Apache configuration solution with XSendFile module. It' really easy to install and use. So I suggest to add in the root option another one like Use XSendFile (true, false)
I just add 
```php
if ($download) array_unshift ($result['header'],'X-Sendfile:' . $volume->realpath($target));
```

before return result in the elFinder,class,php   method File.

Thanks again for your product,
Nikita

geekmug:
I'm new to this codebase, so I did my best to create a patch to resolve #475.

All of the file-based acceleration headers require some amount of configuration on the web server to allow serving files from a path on the system, so this is disabled by default. Once the server has been configured, this functionality can be enabled by adding the "xsendfile" option to a root volume with the name of the header to use (e.g., "X-Sendfile"). The header is configurable because there are a number of variants in the wild, and there is no portable way to know what header should be used.

I have tested this with Apache 2.4 w/ mod_xsendfile 0.12.

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -1526,13 +1526,21 @@ class elFinder {
 				'Connection: close'
 			)
 		);
-
-
-		$xsendfile = $volume->options($target)['xsendfile'];
-		$realpath = $volume->realpath($target);
-		if ($xsendfile !== '' && $realpath !== false) {
-			$result['header'][] = $xsendfile . ': ' . $realpath;
-		} else if (isset($file['url']) && $file['url'] && $file['url'] != 1) {
+		
+		// check 'xsendfile'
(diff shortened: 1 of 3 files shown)


nao-pon:
@geekmug Thank you your contribute. I corrected handling of an option `xsendfile`.

Thanks again! 👍

geekmug:
Ah wow, thanks for handling the details that I didn't know about! I appreciate your efforts!

nao-pon:
@geekmug Yes, that's light. I'll fix it. ;-)

PR #2093, opened 2017-07-07
Fixed error: Using $this when not in object context

alexantr:
Fixed fatal error when trying to call `elFinder::getApiFullVersion()`.
I found only one way: convert `$version` and `$revision` to static vars.

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -14,16 +14,24 @@ class elFinder {
 	/**
 	 * API version number
 	 *
+	 * @var float
+	 **/
+	protected static $ApiVersion = 2.1;
+	
+	/**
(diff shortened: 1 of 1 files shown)


php/elFinder.class.php
```diff
@@ -1273,7 +1273,7 @@ protected function open($args) {
 		}
 		
 		if (!empty($args['init'])) {
-			$result['api'] = $this->version + ($this->revision / 10000);
+			$result['api'] = (float)self::$version + (self::$revision / 10000);
```
alexantr: Deal with inspection in IDE

nao-pon:
@alexantr Thanks! 👍

PR #2100, opened 2017-07-13
More french translations.

mducharme:
Created with https://studio-42.github.io/elFinder/tools/langman/#2.1:fr


nao-pon:
@mducharme Thanks! Sorry for being late.

Issue #2109, opened 2017-07-23
elFinder Ver: 2.1.26


Chaffinch:
Hi Naoki
On:
https://github.com/Chaffinch/elFinder/tree/master/js/i18n

I added: translation for >>> elFinder Ver: 2.1.26

PR #2110, opened 2017-07-24
[i18n:pl] update to 2.1.26

nao-pon:


nao-pon:
@Chaffinch Thanks! 👍 closes #2109

Issue #2156, opened 2017-08-19
elFinder/js/i18n/elfinder.pl.js


Chaffinch:
elFinder/js/i18n/elfinder.pl.js
has been updated to version 2.1.28 - 2017-08-15
https://github.com/Chaffinch/elFinder/commit/8011db4d55716c7820596b049511125e6c2cc6c1
👍

PR #2157, opened 2017-08-19
[i18n:pl] elfinder.pl.js version2017-08-16

nao-pon:
[i18n:LANG] add "cmdselectall", "cmdselectnone", "cmdselectinvert"


nao-pon:
@Chaffinch Thanks! also closes #2156

PR #2178, opened 2017-08-30
[js:core] add an option `fileFilter` to make filtering on browsing

Flamenco:
**Documentation To Be Added**
To filter directory results by filename, set the filenameFilter option to a valid regular expression.

If the filter evaluates to false, no filtering will occur (null, undefined, false, etc)
If the filter is not a RegExp, or an invalid RegExp, no files will pass the filter.
Directory Names are not filtered.

**Examples**
```javascript
{
    // pdf files
    filenameFilter = /.*\.pdf$/i
}
// or
{
    // png and jpg files
    filenameFilter = /.*\.(png|jpg)$/i
}
// or
{
    // no hidden files 
    filenameFilter = /[^\.].*/
}
```

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -6356,6 +6356,21 @@ elFinder.prototype = {
 	 */
 	normalize : function(data) {
 		var self   = this,
+			fileFilter = (function() {
+				var func, filter;
+				if (filter = self.options.fileFilter) {
+					if (typeof filter === 'function') {
+						func = function(file) {
+							return filter.call(self, file);
+						}
+					} else if (filter instanceof RegExp) {
+						func = function(file) {
+							return filter.test(file.name);
+						}
+					}
+				}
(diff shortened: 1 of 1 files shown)


nao-pon:
@Flamenco Thanks! I edited to change option name `filenameFilter` to `fileFilter` and accept function.

nao-pon:
**Options example**
```javascript
{
    // pdf files
    fileFilter = /.*\.pdf$/i
}
// or
{
    // png and jpg files
    fileFilter = /.*\.(png|jpg)$/i
}
// or
{
    // no hidden files 
    fileFilter = /[^\.].*/
}
// or
{
    // check MIME-type
    fileFilter = function(file) {
        var self = this; // `this` is instance of elFinder object
        return file.mime && file.mime.match(/^image\//i); image only
    }
}
```

Issue #2194, opened 2017-09-06
Ftp server with Virtual users and the ftp owner configuration option


arubalucalasagni:
Many ftp servers can be accessed by a **Virtual User**.
For Example with [pure-ftpd](https://www.pureftpd.org/project/pure-ftpd) you can configure an ftp user that it isn't a system user. In that case the user logged is different from the owner of the files but the virtual user could have the write permissions, i.e. it could be menage the file like the real owner, see [Virtual user](https://download.pureftpd.org/pub/pure-ftpd/doc/README.Virtual-Users).
We spotted some bugs when we tried to use the Ftp Drive with a Virtual User.
To show this problems I have prepared a ftp server to be accessed both a System User and a Virtual User that has the same permissions (it is like an alias of the owner) and after I have configured a connector with four ftp drivers, changing the owner option and creating a lot of files with different permissions and owners.

### 1. Ftp Driver, Virtual User, option owner = true
![ftpdrivervirtualuserownertrue2](https://user-images.githubusercontent.com/24244107/30116586-3bc62b7e-931e-11e7-80c7-9d3dcf9a21af.png)
Here we don't have the write permissions for the owner files but actually we could write them. 
Oddly it seems that we have write permissions for this two root objects but it's not true:
1. F755root.txt
2. F644root.txt
  
### 2. Ftp Driver, Virtual User, option owner = false
![ftpdrivervirtualuserownerfalse2](https://user-images.githubusercontent.com/24244107/30116948-23401ea6-931f-11e7-8195-d9c58685f36d.png)
Here we have the same problems with the owner files, we could write them but we cannot. 
Now the problem with the root files is disappeared.
There is a little green icon in any item. 

### 3. Ftp Driver, user Owner, option owner = true
![ftpdriverrealuserownertrue2](https://user-images.githubusercontent.com/24244107/30117406-6672310e-9320-11e7-8a9a-1cf812a94e34.png)
Here is the best configuration but it's not always possible use a system user to make an ftp access.
There is the same bug with the first configuration:
1. F755root.txt
2. F644root.txt

### 4. Ftp Driver, user Owner, option owner = false
![ftpdriverrealuserownerfalse3](https://user-images.githubusercontent.com/24244107/30118209-b9decab2-9322-11e7-9f56-926968dce1a6.png)
Here we aren't able to create new files or folders even if we are in a folder with write and read permissions nevertheless we can edit or rename the existing files.
There is a little green icon in any item. 

[...]

PR #2195, opened 2017-09-07
Ftp server with Virtual users and the ftp owner configuration option

arubalucalasagni:
Here is the Pull Request to resolve this issue
#2194 
Now we consider the option owner in this way:
1. If the option owner is True it means that we don't check if we are the system owner of the object, elFinder will always considerer you the owner. This is the same default behaviour of other ftp clients and it's the only way to use Virtual users.
2. If the option owner is False or Empty Elfinder will compare the file system user with the logged user and so if the user of the file is equal to the ftp user, the logged user will be the owner otherwise it will be an other user.
So if the logged in ftp user is a Virtual User we could set the option owner to True, if the logged in ftp user is also a system user we can set the owner option to False.

We could add this usage for the option owner here https://github.com/Studio-42/elFinder/wiki/Connector-configuration-options-2.1
Actually its default value is True. This means that when we mount a new FTP Driver using the interface with this fix we will have the combination Virtual User with owner = True or System User with owner = true. So in the last case ElFinder will not compare the logged in ftp user with the file owner, nevertheless the user is able to make any action for both cases. If the virtual user cannot change the file the right message is already shown. 

To resolve this we could add a check box to say if the user is a Virtual user in the __Mount Network Volume__ when the ftp is selected.

Below we can see the same cases of the issue but with the fix applied:

### 1. Ftp Driver, Virtual User, option owner = true
ftpdrivervirtualuserownertrue
Now the user can work adding editing and deleting files. The drawback is if a file is of an other user and we try to write we will get an error message, see this two files:
1. F755root.txt
2. F644root.txt

### 2. Ftp Driver, Virtual User, option owner = false
ftpdrivervirtualuserownerfalse
This combination it's not good.

### 3. Ftp Driver, user Owner, option owner = true
ftpdriverrealuserownertrue
The user can work adding editing and deleting files. The drawback is if a file is of an other user and we try to write we will get an error message, see this two files:
1. F755root.txt
2. F644root.txt

### 4. Ftp Driver, user Owner, option owner = false
ftpdriverrealuserownerfalse
I don't see any problem, the best combination.

### DIFF
--- a/php/elFinderVolumeFTP.class.php
+++ b/php/elFinderVolumeFTP.class.php
@@ -406,11 +406,18 @@ class elFinderVolumeFTP extends elFinderVolumeDriver {
 			$stat['owner'] = $info[2];
 			$stat['group'] = $info[3];
 			$stat['perm']  = substr($info[0], 1);
-			$stat['isowner'] = $stat['owner']? ($stat['owner'] == $this->options['user']) : $this->options['owner'];
+                	//
+			// if not exists owner in LS ftp ==>                    isowner = True
+                	// if is defined as option : 'owner' => true            isowner = True
+                	// 
+                	// if exist owner in LS ftp  and 'owner' => False        isowner =   result of    owner(file) == user(logged with ftp)
+                        //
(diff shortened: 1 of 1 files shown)


nao-pon:
@arubalucalasagni Thanks! 👍

Issue #2205, opened 2017-09-17
v.2.1 /tmp openbasedir error


stalker780:
Many shared hostings have problems with openbasedir and /tmp folder

WARNING: finfo_file(): open_basedir restriction in effect. File(/tmp/phpoIyUCq) is not within the allowed path(s): (/var/www/website/data:.) in elFinder.class.php line 2123.

WARNING: finfo_file(/tmp/phpoIyUCq): failed to open stream: Operation not permitted in elFinder.class.php line 2123.

WARNING: file_get_contents(): open_basedir restriction in effect. File(/tmp/phpoIyUCq) is not within the allowed path(s): (/var/www/website/data:.) in elFinder.class.php line 2891.

WARNING: file_get_contents(/tmp/phpoIyUCq): failed to open stream: Operation not permitted in elFinder.class.php line 2891.

WARNING: is_readable(): open_basedir restriction in effect. File(/tmp/phpQNdVDX) is not within the allowed path(s): (/var/www/website/data:.) in elFinderVolumeDriver.class.php line 4195.

This happens when I try to resize an image and press Save As.

Is it possible to use commonTempPath instead of sys_get_temp_dir() and tmpfile()?

nao-pon:
@stalker780 Thanks! 👍

PR #2206, opened 2017-09-18
[php:core] use `fopen($volume->getTempFile(), 'wb')` instead `tmpfile()`

stalker780:
use getTempDir instead of system tmp path, to prevent an error when tmp is not in open_basedir list
WARNING: finfo_file(): open_basedir restriction in effect. File(/tmp/phpoIyUCq) is not within the allowed path(s): (/var/www/website/data:.) in elFinder.class.php line 2123.
#2205


nao-pon:
@stalker780 Thanks! 👍

PR #2207, opened 2017-09-18
[VD:abstract] use `fopen($this->getTempFile(),'wb')` instead `tmpfile()`

stalker780:
use getTempPath instead of system tmp path, to prevent an error when tmp is not in open_basedir list
WARNING: is_readable(): open_basedir restriction in effect. File(/tmp/phpQNdVDX) is not within the allowed path(s): (/var/www/website/data:.) in elFinderVolumeDriver.class.php line 4195.
#2205


nao-pon:
@stalker780 Thanks! 👍

PR #2208, opened 2017-09-18

stalker780:
Uncaught Error: Call to protected method elFinderVolumeDriver::getTempFile() from context 'elFinder'


nao-pon:
@stalker780 Changing the design of existing methods can cause problems if someone is using inheritance classes. I fixed this problem with 84ac1c51ba298920106d5ef9e0806f9a1c5d4d09.

Thank you so much!

Issue #2212, opened 2017-09-21
Using epsv4 off with Pure-ftpd servers creates a timeout


arubalucalasagni:
In elFinderVolumeFTP.class.php there is this instruction in the connect() method:
```php
ftp_raw($this->connect, 'epsv4 off' );
```
I don't know the purpose of that line but if it is utilized with a pure-ftpd server it creates a timeout:
```cmd
ftp> remotehelp
214-The following SITE commands are recognized
 ALIAS
 CHMOD
 IDLE
 UTIME
214 Pure-FTPd - http://pureftpd.org/
ftp> literal epsv4 off
Connessione chiusa dall'host remoto.
```
It's not a big problem because the other commands work well, but every time that we pass through that line we waste time.

The reality is that the real bug is present in pure-ftpd: if you send a server command that ends with a number, for example rewqr12345, the server doesn't reply with a "500 Unknow Command" but it doesn't reply at all and the remote host close the connection.

We observed the timeout with this pure-ftpd versions:
1.0.43 (its response is above)
1.0.46
![image](https://user-images.githubusercontent.com/24244107/30685046-cdea3f9a-9eb3-11e7-8add-9da0eea8d38f.png)

With a previous version 1.0.36 the server gave us an answer:
![image](https://user-images.githubusercontent.com/24244107/30685056-d964f48c-9eb3-11e7-9b89-c7f571556ad4.png)

We are working to this problem carrying out this row only if the server isn't a pure-ftpd server.
We will make you a Pull Request as soon as possible.

Luca

PR #2213, opened 2017-09-21
Fix Using epsv4 off with Pure-ftpd servers creates a timeout

arubalucalasagni:
Here is the fix to remove the client timeout when the ftp server is pure-ftpd, like is explained in the issue #2212.

Maybe we could think to remove this line:
```php
ftp_raw($this->connect, 'epsv4 off' );
```
but I don't know why it was introduced and what kind of ftp servers could understand the command.
The problem, however, is only for the current versions of pure-ftpd, others servers should answer:
```cmd
500 Unknown Command
```
Luca


nao-pon:
@arubalucalasagni Thanks! 👍 

Also closes #2212

Issue #2226, opened 2017-09-27
False errors using the LocalFileSystem Driver and working with empty files


arubalucalasagni:
Using a LocalFileSystem Volume Driver if you open an empty text file with an editor and you try to save the file you achieve this error "Unable to save" even if the file is correctly saved:

![image](https://user-images.githubusercontent.com/24244107/30907656-82379a26-a37b-11e7-8e53-1f157383fdab.png)

You can replicate the bug also taking a text file, removing all content and save the file. In the below example I made a save as:
![image](https://user-images.githubusercontent.com/24244107/30908251-527c5b9e-a37d-11e7-9ddb-7966b6569107.png)
Even in this case the file is saved, so the bug is the prompted error.

We are going to prepare a fix for this problem.

nao-pon:
@arubalucalasagni Probably this problem was fixed by #2207 (cf. #2205). Please try and check with [nightly build](https://github.com/Studio-42/elFinder/archive/2.1.zip).

arubalucalasagni:
@nao-pon The test was done with an aligned master, personally I did a rebase, today, and the master has already these commits.

For Example here https://hypweb.net/elFinder-nightly/demo/2.1/ you can create a file, open the file with a text editor and then try to save the empty file. If the file has some content you have to remove the content.
![image](https://user-images.githubusercontent.com/24244107/30925907-bc96ebbc-a3b3-11e7-9a85-db919bc82e22.png)

CodeLyokoXtEAM:
Confirm it present in demo site
![screenshot_20170927-224521](https://user-images.githubusercontent.com/18566890/30927284-9e133700-a3d5-11e7-903d-dbd87aa6fab1.png)

nao-pon:
@arubalucalasagni Yes, you are right. Thank you so much. 👍

nao-pon:
@CodeLyokoXtEAM Thanks! 👍

PR #2227, opened 2017-09-27
False errors using the LocalFileSystem Driver and working with empty

arubalucalasagni:
Fix for this bug #2226
The problem is that the php function file_put_contents returns the number of bytes that were written to the file, or FALSE on failure. 
```php
protected function _filePutContents($path, $content) {
		return file_put_contents($path, $content, LOCK_EX);
	}
```
So the response on  Failure was False and the response for an empty file correctly saved was 0.
I changed the function to return only Boolean value.


nao-pon:
@arubalucalasagni Thanks! 👍

PR #2231, opened 2017-09-30

stalker780:
Make mime error message more informative. Show mime type blocked.
Cause I faced a problem uploading tgz files and spent several hours finding right mime type, which I need to allow.

function elFinder->upload() required application/x-compressed type
and
function elFinderVolumeDriver->upload() required additional type application/x-gzip

But a VolumeDriver showed only an error: File type not allowed.

Now it will show:
File type not allowed.
(application/x-gzip)


nao-pon:
@stalker780 Thanks! 👍

PR #2234, opened 2017-10-02
create zip edit php

CodeLyokoXtEAM:
get file list is must modify(I create it for manually check up zip files). Also request data is shown begin zip.php
Elfinder need ui to control it.
add files/folders is not completed yet.
extract selected file have issue that it's extracted folder structure in zip.
so it must be extracted to temp folder and after move files.

'Delete folders/files, rename folders/files, open files inside zip, create folder inside zip' is separately call in code.
Edited:-
extract function can unzip zipcrypto encrypted zip.
open files doesn't work with encrypted zip.so user must unzip selected files using password and open it.

Also it must modify input data fields to combine to elfinder. I haven't get any clue to doing it.

### DIFF
--- a/php/plugins/archive/zip.php
+++ b/php/plugins/archive/zip.php
@@ -1,10 +1,38 @@
 <?php
 
 //request data
-$zip_path='C:/wamp64/www/files/files.zip';//zip file path
+//$zip_path='';//zip file path
+
+if (isset($_GET["zip"])) {
+	$zip_path=$_GET["zip"];
+}
+else {
+	echo 'Zip Path Not Given';
+	die();
+};
 $pass='123';//user password. Its can only use for extract ZipCrypto Encryption zip
-$cmd='get';//'get','rename','delete','open','create','add', 'extract'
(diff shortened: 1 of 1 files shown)


nao-pon:
@CodeLyokoXtEAM Thank you for your work. 

The UI is essential for elFinder. We can not accept this PR until UI is implemented. I thought. There is an adapter that handles [ZIP in FlySystem](http://flysystem.thephpleague.com/adapter/zip-archive/), so there is a method to make it network mount using it.

CodeLyokoXtEAM:
I add simple ui. Not Working all functions. only browsing files & download files (Not folders).
Zip archive Function only work local files.that mean only inside server(not Remote Url.if You want to open remote url You must Copy zip to temp folder which inside server. I didn't complete that part doing automatically).

to open Zip 'http://localhost/files/zip.php?&zip="Your Zip file relative Path"'
ex:-
'http://localhost/files/zip.php?&zip=./files.zip'

I found Some Bug in code. so File Browsing inside may be wiard.

image

CodeLyokoXtEAM:
Bracket Bug fixed :-)

PR #2267, opened 2017-11-04
Make help links in one line

alexantr:
Propose to set help links in one line.

Before:

sclrolh

After:

uyztjjo

Compact view:

c2ywptl

RTL:

esjhr1n


nao-pon:
@alexantr Thanks! 👍

PR #2272, opened 2017-11-09
Update mime.types according new font mimes

alexantr:
Got updated mime.types from apache trunk:
https://svn.apache.org/repos/asf/httpd/httpd/trunk/docs/conf/mime.types

Font mime on IANA:
https://www.iana.org/assignments/media-types/media-types.xhtml#font


nao-pon:
@alexantr Thanks! 👍

PR #2283, opened 2017-11-16
Fix: thumbnails not displayed with newAPI

pcable:
If file.tmb is populated but there is no tmbURL specified then the thumbnail will not be rendered.


nao-pon:
@pcable Thanks! 👍

PR #2289, opened 2017-11-21
Add support for FTP+SSL

zzplab:
In case of SSL use php function: ftp_ssl_connect(...)
Setting the driver options with SSL:
```
array(
    'driver'        => 'FTP',
    'host'          => 'localhost',
    'user'          => 'eluser',
    'pass'          => 'elpass',
    'port'          => 21,
    'ssl'           => true, // true or false, default: false
    'mode'          => 'passive',
    'path'          => '/',
    'timeout'       => 10,
    'owner'         => true,
    'tmbPath'       => '',
    'tmpPath'       => '',
    'dirMode'       => 0755,
    'fileMode'      => 0644
)
```


nao-pon:
@zzplab Thanks! 👍

PR #2316, opened 2017-12-06
translate and add Korean help

badsaarow:
translated en.html.js to ko.html.js in Korean

Without this file, there was 404 page not found error happened, if set language to Korean and select help.


nao-pon:
@badsaarow Thanks! 👍

PR #2318, opened 2017-12-07
translate untranslated to Korean

badsaarow:
I'd missed some sentence with untranslated on previous PR.
Now whole sentences were translated.


nao-pon:
@badsaarow Thanks! 👍

PR #2323, opened 2017-12-11
Add Ukrainian translation

a2exfr:
Add Ukrainian translation


nao-pon:
@a2exfr I'd like to merge this, but it is closed by you. Do you want to cancel this pull request?

a2exfr:
Yes, cancel please.  I didn't noticed that ua language  is exists like uk language. So I will update uk.

nao-pon:
P.S I found the answer. The country code is ua and the language code is uk.

PR #2346, opened 2018-01-03
rename update event to previewupdate because of conflict with prototype

vcordes79:
I still need prototype js for my app and could not use elfinder because of a conflict with the name of the update event. So I renamed the event to previewupdate. Please integrate the change into mainstream version.
Also I added an option to use ViewerJS as preview for pdf files in quicklook and fixed the quicklook close / dock button alignment

### DIFF
--- a/css/quicklook.css
+++ b/css/quicklook.css
@@ -122,7 +122,8 @@
 }
 .elfinder-quicklook-titlebar-icon .ui-icon {
 	position: relative;
-	margin: -9px 3px 0px 0px;
+	margin: 0px 3px 0px 0px;
+	display: inline-block;
 	cursor: pointer;
 	border-radius: 10px;
 	border: 1px solid;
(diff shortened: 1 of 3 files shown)


css/quicklook.css
```diff
@@ -122,7 +122,8 @@
 }
 .elfinder-quicklook-titlebar-icon .ui-icon {
 	position: relative;
-	margin: -9px 3px 0px 0px;
+	margin: 0px 3px 0px 0px;
+	display: inline-block;
```
nao-pon: What kind of problem are you having? Is there a screenshot of the problem?

nao-pon:
@vcordes79 The event name "update" can not be changed easily because there are cases where other users have added the quicklook preview plugin by themselves. Instead, I will modify it to change the event name if `Element.update` is extended.

Thanks!

css/quicklook.css
nao-pon: When I checked, the display position shifts downward due to this change.

nao-pon: If your amendment is valid for plain elFinder, please make another branch and send it as a separate pull request.

PR #2374, opened 2018-01-26
All translation of Chinese

china-live:
qq 20180126195800
qq 20180126195833
qq 20180126195937


nao-pon:
@china-live Thanks! 👍

PR #2387, opened 2018-02-16
[ENH] Lossless image compression. Saved 1kb with an average of 35.7% …

drsassafras:
[ENH] Lossless image compression. Saved 1kb with an average of 35.7% per file, up to 67.7%.


nao-pon:
@drsassafras Thanks! 👍

PR #2432, opened 2018-03-21
call back function or method on sanitizeFileName

Yves-K:
sanitizeFileName is great but we can't do everything with arrays or regular expressions...

I propose to add a callBack param on plugin config so we can use a function (or method) to fine tune the filename.

Tested on multiple production sites with success


nao-pon:
@Yves-K Thanks! 👍

PR #2442, opened 2018-03-26
Add option onDropWith for plugin function iaEnabled

nmpetkov:
It is handy in some cases plugins (AutoResize, AutoRotate, Watermark) to be enabled, but to function only if  given meta key is pressed while dropping images. This is opposite to option offDropWith.

This PR adds such a possibility.

Current functionality (offDropWith) is same and new features are backward compatible.


nao-pon:
@nmpetkov It's so nice idea! Thanks! 👍

PR #2449, opened 2018-03-29
Fix stream reading order to stderr then stdout so it doesn't hang 

ioweb-gr:
I had a case where elfinder would hang in Open Folder indefinitely or until timeout. This was caused by the reading order for streams.

Currently elfinder is reading stdout before stderr
There is a case where stdout will never return after an error while elfinder is still waiting for it.

I found the solution on stackoverflow for my case 
https://stackoverflow.com/questions/31194152/proc-open-hangs-when-trying-to-read-from-a-stream


nao-pon:
@ioweb-gr Thanks! 👍

PR #2481, opened 2018-04-14
Bind listener after upload save

Yves-K:
Hello,

Working on many plugins projects for elFinder it came necessary to acces uploaded data before returning to the client.

For instance, i created a plugin that always create a .jpeg clone of non jpeg files to obtain an optimized version.

So i suggest to bind plugins on after ulpoad just passing the results array and elFinder instance.

Regards

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -3013,6 +3013,13 @@ class elFinder {
 				'bind' => 'upload'
 			);
 		}
+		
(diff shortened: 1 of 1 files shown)


nao-pon:
@Yves-K Did not you satisfy your requirements with the `upload` event listener?

Yves-K:
Sorry, i did not notice commands listeners.
Yes it's perfect ;-)

PR #2533, opened 2018-06-06
Update french translation

Jonathan57500:


nao-pon:
@Jonathan57500 Thanks! 👍

PR #2544, opened 2018-06-12
 [VD:abstract] fallback imgLib to 'gd' when error in Imagick

### LINKED ISSUE #2540: stopped generating preview images
I see that EF stopped generate preview images.
Only old images created in the past are exists. New are like default icon only.
I checked function createTmb in elFinderVolumeDriver
and
imagick throw error: NoDecodeDelegateForThisImageFormat `JPEG' @ error/constitute.c/ReadImage/501
But nowhere is this information.
I disabled imagick and EF uses gd but preview images are created.

nao-pon:
To fix #2540 .


nao-pon:
@standus I corrected my mistake. Thanks!

PR #2556, opened 2018-06-20
The search doesn't retrieve the files when the name of the file starts with the searched word.

unigia:
The search doesn't retrieve the files when the name of the file starts with the searched word, only for the driver like FTP that doesn't have a doSearch implementation.
For example when we call the method "searchMatchName" If the parameter $query is equal to $name the conversion of 0 into bool returns erroneously false.
This PR resolve it. 
Thanks.


nao-pon:
@unigia Thank you so much! 👍

PR #2574, opened 2018-07-06
openoffice(libreoffice) files open in MS Viewer

jacceko:
MS Viewer + .odt .ods. odp -  it work :-)


nao-pon:
@jacceko Thanks! 👍

PR #2613, opened 2018-08-10
fixed filesize and filemtime bug

yuchting:
filesize problem for file larger than 4GB ( testing in PHP Version 5.4.16)
https://stackoverflow.com/questions/6482211/php-filesize-over-4gb

and

sometimes filemtime will failed and return false

image

### DIFF
--- a/php/elFinderVolumeLocalFileSystem.class.php
+++ b/php/elFinderVolumeLocalFileSystem.class.php
@@ -488,6 +488,17 @@ class elFinderVolumeLocalFileSystem extends elFinderVolumeDriver {
 		if ($readable) {
 			$size = sprintf('%u', filesize($path));
 			$stat['ts'] = filemtime($path);
+			if($size == '0'){
+				// filesize problem for file larger than 4GB ( testing in PHP Version 5.4.16)
+				// https://stackoverflow.com/questions/6482211/php-filesize-over-4gb
+				$size = trim(shell_exec('stat -c %s '.escapeshellarg($path)));
+			}
+            
+			if($stat['ts'] === false){
+				// sometime filemtime will failed to return false
+				$stat['ts'] = (int)trim(shell_exec('stat -c %Y '.escapeshellarg($path)));
+			}
+			
 			if ($this->statOwner) {
 				$fstat = stat($path);
 				$uid = $fstat['uid'];


yuchting:
image

nao-pon:
@yuchting The suggested code is valid only on * nix. Can you respond to other environments?

yuchting:
Do you want another information?

nao-pon:
I done in f05b84b0a6166fce2a55a2722b23ee6a8566088c .

yuchting:
@nao-pon sorry for delay reply, actually, I find the [PHP official document](http://php.net/manual/en/function.filesize.php) about "filesize" says: 

> Note: Because PHP's integer type is signed and many platforms use 32bit integers, some filesystem functions may return unexpected results for files which are larger than 2GB.

so I am very sure you need change something in this place for get file size more robustness and then pull this request.

I am not good at PHP and just use this great tool before few weeks ago, I don't know if I change it well or not, just with a coder habit: found a bug, pull a request.

I am so glad to see you change this and close this PR, thanks!

nao-pon:
@yuchting The best solution to this problem is to use PHP compiled with 64 bit. However, even with the current code, file size of 4 GB or less can be acquired correctly in many environments. I added this option for environments that can not handle files larger than 4GB or 2GB correctly. For example, to realize the same thing as this PR, specify the option as follows.

```php
$opts = array(
    'locale' => '',
    'roots'  => array(
        array(
            'driver' => 'LocalFileSystem',
            'path'   => '/path/to/files/',
            'statCorrector' => function (&$stat, $path, $statOwner, $volumeDriveInstance) {
                if (isset($stat['size']) && $stat['mime'] !== 'directory') {
                    if ($stat['size'] == '0') {
                        // filesize problem for file larger than 4GB ( testing in PHP Version 5.4.16)
                        // https://stackoverflow.com/questions/6482211/php-filesize-over-4gb
                        $stat['size'] = trim(shell_exec('stat -c %s '.escapeshellarg($path)));
                    }
                    
                    if ($stat['ts'] === false){
                        // sometime filemtime will failed to return false
                        $stat['ts'] = (int)trim(shell_exec('stat -c %Y '.escapeshellarg($path)));
                    }
                }
            }
        )
    )
);

// run elFinder
$connector = new elFinderConnector(new elFinder($opts));
$connector->run();
```

PR #2619, opened 2018-08-17
[VD:Dropbox2] fix options definition test

flohw:
The empty test an undefined variable. I think it should test the `options` property instead.

Updated it in a symfony project, it works a lot better :wink:


nao-pon:
@flohw Oh! Thank you so much! 👍

PR #2696, opened 2018-10-07
check manual config for archivers before auto detect in getArchivers() function

seekerliu:
In `elFinderVolumeDriver.class.php` says user can custom archivers in config, but the `getArchivers()` function did not check the manual config.


nao-pon:
@seekerliu Thanks! 👍

PR #2802, opened 2018-12-17
Broken Handling of Certain Hash Characters from elFinder Connector

mtran0011:
In the elFinder [docs](https://github.com/Studio-42/elFinder/wiki/Client-Server-API-2.1#file-paths-hashes), it has the below content:

> Using this algorithm even without encryption, client cannot get real file paths on the server only relative to root paths. This hash algorithm is recommended but you can use your own implementation so long as it matches these 2 rules:
> 
> hash must be valid for storage in the id attribute of an HTML tag
> hash must be reversible by connector

This doesn't actually hold as characters such as ".", ":" are [valid HTML ID characters](https://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html), but when used in a JQuery hash selector, elFinder would behave strangely as internally, it does it in such a way where these characters are confused with CSS/JQuery selector characters instead of being part of the HTML ID.

For more information on the above issue, refer to the accepted answer on this [stackoverflow post](https://stackoverflow.com/questions/70579/what-are-valid-values-for-the-id-attribute-in-html).

As such, I've made a pull request in an attempt to solve the above issue. So essentially, instead of:
```
$("#" + elementId)
```
It now uses:
```
$(document.getElementById(elementId))
```

**Benchmark**

I ran a small benchmark to see any performance difference, and it turns out using $(document.getElementById("<ElementId>")) is actually faster.

I copied the code within this text file and ran it on the [elfinder demo page](https://studio-42.github.io/elFinder/#elf_l1_Lw).
[Benchmark.txt](https://github.com/Studio-42/elFinder/files/2684138/Benchmark.txt)

**Benchmark Results**

benchmark

### DIFF
--- a/js/commands/download.js
+++ b/js/commands/download.js
@@ -39,7 +39,7 @@ elFinder.prototype.commands.download = function() {
 					var f = fm.file(h),
 						res = (! f || (! czipdl && f.mime === 'directory') || ! fm.isCommandEnabled(mixedCmd, h))? false : true;
 					if (f && inExec && ! res) {
-						$('#' + fm.cwdHash2Id(f.hash)).trigger('unselect');
+						$(document.getElementById(fm.cwdHash2Id(f.hash))).trigger('unselect');
 					}
 					return res;
 				});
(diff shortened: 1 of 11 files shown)


nao-pon:
@mtran0011 Thank you for accurate PR. 👍

PR #2874, opened 2019-03-19
Update elfinder.it.js

danielebruni:
Better (I think) translation for the "places" function resources.


nao-pon:
@danielebruni Thanks! 👍

PR #2933, opened 2019-06-11
Update elfinder.fr.js

fredericauer:
Replaced "places" with "favoris" which is more usual in French.


nao-pon:
@fredericauer Thanks! 👍

PR #2950, opened 2019-06-28
No way to execute this part of code

sashsvamir:
I'm can't find case where this part of code will be executed. This code should be executed only if `$args['target']` is non empty array, but in that case on line 1029 behind variable `$dst` will be set, so line 1032 will be skipped.
Sorry if i'm missed something there.


nao-pon:
@sashsvamir Thank you for contributing to this project.

Line 1029 deals with 'target' and line 1032 deals with 'targets'. They are different parameters.

So I close this. Thank you!

sashsvamir:
oh, i'm sorry, I'll be more attentive next time

nao-pon:
@sashsvamir Don’t worry about it. We always welcome any PR! 👍

PR #2952, opened 2019-06-28
fix curl params

sashsvamir:
You should enable in curl follow rediects, otherwise script may fail


nao-pon:
@sashsvamir Thanks! 👍

PR #3008, opened 2019-10-01
Update cwd.css

dbugit:
Text position on opendocument icon corrected

### DIFF
--- a/css/cwd.css
+++ b/css/cwd.css
@@ -533,7 +533,7 @@
     min-width: 20px;
     max-width: 84px;
     text-align: center;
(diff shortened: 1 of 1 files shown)


nao-pon:
@dbugit Thank you for contributing to the elFinder project.

I understand that the purpose of this PR is to adjust the minimum width of the type badge. However, adjustment by padding has an adverse effect on badges with many characters. So I adjusted the minimum width.

Thanks! 👍

nao-pon:
@dbugit It was the influence of box sizing by third party CSS. I re-corrected it.

PR #3009, opened 2019-10-03
A option to avoid the command "open in new windows"

w1p30ut3r:
A option to avoid the command "open in new window"... To avoid open new windows from a folder for example... It would be great! Or there is a way right now? :S


nao-pon:
@w1p30ut3r I think this is a wrong pull request, so close it.

PR #3028, opened 2019-11-04
Fix CustomData to Thumbnail URL

ebonato:
Sorry about not providing a full use case for this issue, but its clear to see..
If you use customData on client settings, customData params is only used when fetching data from connector URL. 
If you have a usecase where one or more of these customData params defines the root of connector, thumbnail URL also needs these params to provide correct thumbnail.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -2111,12 +2111,12 @@ var elFinder = function(elm, opts, bootCallback) {
 			}
 			if (url) {
 				if (file.ts && tmbUrl !== 'self') {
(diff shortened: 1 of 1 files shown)


nao-pon:
@ebonato I add a volume option `tmbReqCustomData` that adds custom data to the thumbnail.

nao-pon:
@ebonato I done. It need an option `'tmbReqCustomData' => true`.

```php
$opts = array(
    'roots'  => array(
        array(
            'driver' => 'LocalFileSystem',
            'path'   => '/path/to/files/',
            'URL'    => 'http://localhost/to/files/',
            'tmbReqCustomData' => true
        )
    )
);
```

Issue #3033, opened 2019-11-19
Required for Google Drive network mount with Flysystem with composer. Installation failed


migrad:
Hello,

With 2.1.50 when I run:

> composer require nao-pon/flysystem-google-drive:~1.1 nao-pon/elfinder-flysystem-driver-ext

 I get the following result:

> Using version ^1.0 for nao-pon/elfinder-flysystem-driver-ext
> ./composer.json has been updated
> Loading composer repositories with package information
> Updating dependencies (including require-dev)
> Your requirements could not be resolved to an installable set of packages.
> 
>   Problem 1
>     - nao-pon/elfinder-flysystem-driver-ext 1.0.0 requires barryvdh/elfinder-flysystem-driver ^0.1.5|^0.2 -> satisfiable by barryvdh/elfinder-flysystem-driver[v0.1.5, v0.2.0, v0.2.1].
>     - nao-pon/elfinder-flysystem-driver-ext 1.0.1 requires barryvdh/elfinder-flysystem-driver ^0.1.5|^0.2 -> satisfiable by barryvdh/elfinder-flysystem-driver[v0.1.5, v0.2.0, v0.2.1].
>     - nao-pon/elfinder-flysystem-driver-ext 1.0.2 requires barryvdh/elfinder-flysystem-driver ^0.1.5|^0.2 -> satisfiable by barryvdh/elfinder-flysystem-driver[v0.1.5, v0.2.0, v0.2.1].
[...]

PR #3039, opened 2019-11-22
Fix invalid grant bad request

migrad:
Hello @nao-pon,

Regardless of the problem with composer ( #3033 ), I have copied the dependencies of a previous version, and it works correctly.

However, when I try to connect to Google Drive it gives me "errReauthRequire". I have located that this error occurs because authentication failed with the "invalid_grant - Bad Request" error (id_client, secret_id and include autoload are good).

I have corrected it and get the token.

But the whole process stands still in this return (elFinderFlysystemGoogleDriveNetmount.php) :

```
            if (!empty($_GET['code']) && !$aToken) {
                $aToken = $client->fetchAccessTokenWithAuthCode($_GET['code']);
                $options['access_token'] = $aToken['access_token'];
                $this->session->set('GoogleDriveTokens', $aToken['access_token'])->set('GoogleDriveAuthParams', $options);
                $out = array(
                    'node' => $options['id'],
                    'json' => '{"protocol": "googledrive", "mode": "done", "reset": 1}',
                    'bind' => 'netmount'
                );
                return array('exit' => 'callback', 'out' => $out);
            }
```
Once the callback is executed, a page with a "Close this Window" is displayed and does nothing else.

I understand that at this point, the new window should close and reload elFinder to show the Google Drive driver, is that so?

I have also observed that the session data is deleted, when they should be maintained in order to perform that elFinder reload, right?.

I hope your help, thanks.

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -3923,7 +3923,7 @@ class elFinder
             }
             $script .= 'window.close();';
 
-            $out = '<!DOCTYPE html><html><head><meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"><meta http-equiv="Content-Type" content="text/html; charset=utf-8"><script>' . $script . '</script></head><body><a href="#" onlick="window.close();return false;">Close this window</a></body></html>';
+            $out = '<!DOCTYPE html><html><head><meta http-equiv="X-UA-Compatible" content="IE=edge,chrome=1"><meta http-equiv="Content-Type" content="text/html; charset=utf-8"><script>' . $script . '</script></head><body><a href="#" onclick="window.close();return false;">Close this window</a></body></html>';
 
             header('Content-Type: text/html; charset=utf-8');
             header('Content-Length: ' . strlen($out));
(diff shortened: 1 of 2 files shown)


nao-pon:
@migrad I tested of netomount via FlysystemGoogleDriveNetmount. And It was no problems.

Please try and check with the[ nightly build](https://github.com/Studio-42/elFinder/archive/2.1.zip).

migrad:
I have downloaded it and installed the dependencies, but it gives me the following error (clean installation):

> {"exit":true,"error":"errReauthRequire"}

> /php/connector.minimal.php?cmd=netmount&protocol=googledrive&host=1&code=xxx&scope=https://www.googleapis.com/auth/drive

I have deleted the value of the code parameter, but I receive it correctly.

The constants ELFINDER_GOOGLEDRIVE_CLIENTID and ELFINDER_GOOGLEDRIVE_CLIENTSECRET have the correct values. Uncomment "require './vendor/autoload.php';" and "elFinder::$netDrivers['googledrive'] = 'FlysystemGoogleDriveNetmount';"

At this point I detected that this equality was always false:

> $options['user'] === 'init'

With what $aToken is always false.

nao-pon:
@migrad What are your browsers using? Is the result the same when using a Chrome incognito window?

nao-pon:
@migrad Make sure the `php/.tmp` directory exists and writable by PHP.

nao-pon:
@migrad I close this because we can't reproduce the problem. You can submit a new PR at any time. Thank you!

PR #3048, opened 2019-12-09
Update dialog.css

osworx:
elfinder_message_window
Because value of jquery-ui.min.css is used, the icon is displayed over the text (or text under the icon).

### DIFF
--- a/css/dialog.css
+++ b/css/dialog.css
@@ -188,6 +188,7 @@
     -moz-user-select: text;
     -khtml-user-select: text;
     user-select: text;
+    padding-left: 55px !important;
 }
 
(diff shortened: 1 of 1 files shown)


osworx:
Sorry, has unwanted side efects to all further boxes.
Hhave to check more before creating new pull request.

osworx:
okay, why not creating something like
`<div class="elfinder-dialog-text">..here goes the error text</div>`
and assign **padding-left:40px** to this class?

That would look like this:
elfinder_message_window_corrected

nao-pon:
@osworx Please load jQuery-UI CSS before elFinder CSS.

osworx:
Thanks, that was the case not loading the ui-css before.

PR #3070, opened 2019-12-26
Fix PHP7.4 deprecation error

gwaldvogel:
PHP 7.4 deprecated the use of curly braces to access array/string indexes resulting in errors like this:
`ErrorException  : Array and string offset access syntax with curly braces is deprecated`

This change makes the code compatible with PHP7.4.

For more information on the deprecation see: https://wiki.php.net/rfc/deprecate_curly_braces_array_access


nao-pon:
@gwaldvogel Thanks! 👍

daniloobv:
donde agrego esa lineas de codigo??

PR #3072, opened 2019-12-27
Wrong dialog sizing with dialogContained=true

mahotilo:
When dialogContained is true dialog sizing 
syncSize.width = syncSize.height = syncSize.enabled = true; 
works wrong way, because dialogs windows are maximized by height.

### DIFF
--- a/js/ui/dialog.js
+++ b/js/ui/dialog.js
@@ -103,14 +103,10 @@ $.fn.elfinderdialog = function(opts, fm) {
 		opts.allowMinimize = false;
 	}
 	
-	if (fm.options.dialogContained) {
-		syncSize.width = syncSize.height = syncSize.enabled = true;
-	} else {
-		syncSize.width = (opts.maxWidth === 'window');
-		syncSize.height = (opts.maxHeight === 'window');
-		if (syncSize.width || syncSize.height) {
-			syncSize.enabled = true;
-		}
+	syncSize.width = (opts.maxWidth === 'window');
+	syncSize.height = (opts.maxHeight === 'window');
+	if (syncSize.width || syncSize.height) {
+		syncSize.enabled = true;
 	}
 
 	propagationEvents = fm.arrayFlip(opts.propagationEvents, true);


nao-pon:
@mahotilo The elFinder client option `dialogContained` is a setting to prevent the dialog from protruding from the DOM node of elFinder. What problems are you having? Will you explain with screenshots?

mahotilo:
I use elFinder as a part of CMS.

1. With default elFinder settings I had such picture.
image


2. I wanted to prohibit the movement of elFinder's dialogs beyond elFinder boundaries. So I changed "dialogContained" from false to true. (Am I right?)
I succeeded, but some dialogs e.g. Resize&Rotate became oversized.
image


3. The only solution I had founded is to make a changes in `$.fn.elfinderdialog = function(opts, fm) {...}` which I sent in PR. Now things look like this.
image

nao-pon:
@mahotilo OK, This have bugs in some dialogs, which I will fix.

mahotilo:
Thx.
I had tried a last nightly build with your fix and found some problem.  It works for me, but after minimizing and restoring the Resize&Rotzte dialog vertical scrollbar appears.
Can you see this too?

image

image

nao-pon:
@mahotilo No problem with the default theme. Is the CSS of the theme affecting the scrollbar?

mahotilo:
I can see scrollbar and with the default theme too.

image

Caches cleared, private window...

nao-pon:
@mahotilo Is the CSS other than elFinder loaded on the page affecting?

mahotilo:
Yes, this may be because elFinder is integrated to CMS. I have tried to turn off all css and js adjusting elFinder, but I still see the scrollbar after restoring dialog from the minimized state.

If you can not reproduce this bug, let us assume it is related to my settings.

nao-pon:
Yes, I cannot reproduce this problem. I created a simple test case.

- https://jsbin.com/yayoyam/edit?js,output

mahotilo:
I can not find Resize&Rotate command

nao-pon:
Demo content cannot be edited, so select from the Writable Volume Group.

mahotilo:
Yes, I also can not see the described problem in JS bin.
So it is not on your side. Sorry for the false alarm and thank you for your help.

nao-pon:
You are welcome! 😄

PR #3122, opened 2020-03-09
Fix volume names in tree when names have ampersand character encoded as "&amp;"

balaianu:


nao-pon:
@balaianu In HTML, `"&"` must be represented by `"&amp;"` and cannot be excluded from escaping.

PR #3137, opened 2020-03-29
Update elfinder.pl.js

Chaffinch:
Update 29.03.2020


nao-pon:
@Chaffinch Thank you as always! 👍

PR #3140, opened 2020-04-02
[i18n:ko] Update Korean translation.

LI-NA:
Change date format to KS X ISO 8601.
Update korean translation.


nao-pon:
@LI-NA Thanks! 👍

PR #3144, opened 2020-04-07
[i18n:ko] Update Korean translation.

LI-NA:
All translated content has been reviewed and translated again if necessary.
Also fixed several spelling errors.


nao-pon:
@LI-NA Thanks! 👍 Sorry I'm late to notice.

PR #3203, opened 2020-09-13
Update cookieEnabled check

sergiz:
Hello.

This check is a reason for a warning in modern browsers since SameSite attribute is not set. I believe it can be replaced with a call to navigator API.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -5810,14 +5810,7 @@ elFinder.prototype = {
 	 * 
 	 * @type Boolean
 	 */
-	cookieEnabled : (function() {
-		var res = false,
-			test = 'elftest=';
-		document.cookie = test + '1';
-		res = document.cookie.split(test).length === 2;
-		document.cookie = test + ';max-age=0';
-		return res;
-	})(),
+	cookieEnabled : window.navigator.cookieEnabled,
 
 	/**
 	 * Has RequireJS?


667bdrm:
Looks like here are could be issues with IE11

https://stackoverflow.com/questions/4603289/how-to-detect-that-javascript-and-or-cookies-are-disabled
https://stackoverflow.com/questions/42682529/navigator-cookieenabled-not-working-in-ie11

667bdrm (commented):
Looks like here are could be issues with IE11

https://stackoverflow.com/questions/4603289/how-to-detect-that-javascript-and-or-cookies-are-disabled
https://stackoverflow.com/questions/42682529/navigator-cookieenabled-not-working-in-ie11

Have you checked that it works as expected in other browsers?

https://developer.mozilla.org/en-US/docs/Web/API/Navigator/cookieEnabled

sergiz:
I haven't. It works in browsers that users actually use. Market share of IE11 is hardly 1% (and getting lower every day).

If you believe that this 1% justifies adding more bloat to this package, then feel free to close this PR. However, in that case someone has fix missing sameSite attribute which is an actual problem to other 99%.

667bdrm:
Checked in Win8 IE in Edge, IE10 and IE9 compatibility modes. We could add condition block with old detection method if something broken found in future.

PR #3222, opened 2020-11-15
[VD:SFTP] add new SFTP driver, via phpseclib library

sitecode:
Extending the FTP driver class, add SFTP driver to allow mounting SFTP accounts. Requires the phpseclib v1 library to be installed and option phpseclibDir for the driver to be able to include and instantiate an $sftp object. Or optionally specific a callback via option connectCallback that includes phpseclib library and instantiates and returns an $sftp object itself.

### DIFF
--- a/php/elFinderVolumeSFTPphpseclib.class.php
+++ b/php/elFinderVolumeSFTPphpseclib.class.php
@@ -1,0 +1,838 @@
+<?php
+
+/**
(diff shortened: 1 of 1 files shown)


667bdrm:
Thank You!

sergiz:
Why is phpseclib v1 used and not v2?

Issue #3262, opened 2021-03-03
Upload problem


mguinness:
Firstly thanks for this amazing library.  I'm having an issue when I try to upload a file named GitHub.url with the following contents:

```
[InternetShortcut]
URL=https://github.com/
```

It's an Internet Shortcut (.url file) that is created in Windows.  During upload the `kind` variable in the code below is incorrectly set as 'string' instead of 'file'.

https://github.com/Studio-42/elFinder/blob/f9c906d808d1721a62fc2a4fdb38d77c1c1ff229/js/commands/upload.js#L197

In the Chrome browser console below there are two entries, the first from a text file upload and the second from the url file upload.  As you can see the second `DataTransferItemList` has 3 items instead of 1 and with the code above it only looks at the first item.

![tfr](https://user-images.githubusercontent.com/14002844/109879762-de996e80-7c2a-11eb-9d2a-65b0212a591d.PNG)

I'm not sure what the best solution would be, but I think the code should grep for `file` and if not found use the first item?

You can use the fiddle created at https://jsfiddle.net/ghvfdyam/ to reproduce this issue in both Chrome and Edge browsers.

PR #3264, opened 2021-03-04
Upload bug

mguinness:
Fixes https://github.com/Studio-42/elFinder/issues/3262.  I think this could be related to [Issue 155455: item.webkitGetAsEntry().file() reports an empty file type](https://bugs.chromium.org/p/chromium/issues/detail?id=155455).

Fix uses [DataTransfer.types](https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/types) array to establish if transfer is a file.

> An array of the data formats used in the drag operation. Each format is string. If the drag operation included no data, this list will be empty. If any files are included in the drag operation, then one of the types will be the string Files.

### DIFF
--- a/js/commands/upload.js
+++ b/js/commands/upload.js
@@ -194,10 +194,17 @@ elFinder.prototype.commands.upload = function() {
 				data = null,
 				target = e._target || null,
 				trf = e.dataTransfer || null,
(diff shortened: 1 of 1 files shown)


nao-pon:
@mguinness thanks you. I will verify this.

nao-pon:
@mguinness This looked good so I completed the merge. Thanks! 👍

PR #3268, opened 2021-03-08
Added and grouped connectors

mguinness:
Grouped .NET connectors together and added new connector


nao-pon:
@mguinness Thanks! :+1:

PR #3271, opened 2021-03-10
Option to pass url to request function

mguinness:
While investigating `commandsOptions.info.custom` it would be good to allow action to pass `fm.request` a url option to override the default url in options.

https://github.com/Studio-42/elFinder/blob/33eea861cb07196edbde5d9188f65a6ce81be21f/js/elFinder.options.js#L651-L662

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -2323,7 +2323,7 @@ var elFinder = function(elm, opts, bootCallback) {
 			useCache = (opts.options || {}).cache,
 			// request options
 			options = Object.assign({
-				url      : o.url,
+				url      : opts.url || o.url,
 				async    : true,
 				type     : this.requestType,
 				dataType : 'json',


nao-pon:
@mguinness You can use the `opts.options.url` property.
```javascript
fm.request({ 
    data : { cmd : 'desc', target: file.hash }, 
    options: { url : '_YourConnectorUrl_' }
})
```

mguinness:
Ah I see that [Object.assign()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/assign#description) will overwrite with the url from options obect.  Thanks for pointing that out.

> Properties in the target object are overwritten by properties in the sources if they have the same key.

```
options = Object.assign({
  url      : o.url
}, opts.options || {})
```

PR #3275, opened 2021-03-20
replace space with underscore in file name

deKuddus:
In some cases, if there is a space in the file name then the URL is not working. So we need a file URL without space.

### DIFF
--- a/php/elFinderVolumeDriver.class.php
+++ b/php/elFinderVolumeDriver.class.php
@@ -2441,6 +2441,7 @@ abstract class elFinderVolumeDriver
      */
     public function upload($fp, $dst, $name, $tmpname, $hashes = array())
     {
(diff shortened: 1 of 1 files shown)


mguinness:
I don't have an issue with space in filename, but I'm not using PHP connector.

Another option is to use [validName](https://github.com/Studio-42/elFinder/wiki/Client-configuration-options-2.1#validName) option on client or `nameAccepted` on server.

https://github.com/Studio-42/elFinder/blob/33eea861cb07196edbde5d9188f65a6ce81be21f/php/elFinderVolumeDriver.class.php#L3976-L4001

nao-pon:
@deKuddus , Yes, as @mguinness says. Specific requirements can be set as an option. If there are items that cannot be set as options, we will always accept PR regarding the addition of options.

PR #3280, opened 2021-03-25
Enable drag when copy or cut enabled

mguinness:
I had an issue where dragging wasn't enabled when copy command was removed but cut command was still active.

```JS
bootCallback: function (fm) {
    delete fm.commands.copy;
}
```

This change now allows dragging when only cut is active.  I still have a problem when either `shiftKey` or `ctrlKey` modifier is used to allow copy.  I commented out the line below as a workaround, is there a way to check to see if copy command is enabled?

https://github.com/Studio-42/elFinder/blob/33eea861cb07196edbde5d9188f65a6ce81be21f/js/elFinder.js#L5185

Or maybe a [uiOptions](https://github.com/Studio-42/elFinder/wiki/Client-configuration-options-2.1#uiOptions) similar to metakeyDragout for `cwd`?

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -5182,7 +5182,9 @@ var elFinder = function(elm, opts, bootCallback) {
 						if (ctr !== chk) {
 							ctr = chk;
 							if (helper.is(':visible') && helper.data('dropover') && ! helper.data('droped')) {
-								helper.toggleClass('elfinder-drag-helper-plus', helper.data('locked')? true : ctr);
+								if (self.options.enableDragCopy) {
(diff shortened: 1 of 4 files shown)


mguinness:
Added `enableDragCopy` to [options](https://github.com/Studio-42/elFinder/wiki/Client-configuration-options-2.1) and prevent copy when command not enabled.

nao-pon:
@mguinness I have created Issue #3302 to fix this.

nao-pon:
I fixed by https://github.com/Studio-42/elFinder/commit/5eebd92ec1a9815a8e17249d928646cd0c089675 .

PR #3291, opened 2021-05-05

billynoah:
Rewrite logic for to get mimeDetect for clarity and to resolve a small bug.

This line causes a bug where it becomes impossible to set mimeDetect to 'mime_content_type': https://github.com/Studio-42/elFinder/blob/33eea861cb07196edbde5d9188f65a6ce81be21f/php/elFinderVolumeDriver.class.php#L1290

The logic here first makes an array of methods available in the current environment.  Then we check to see if the type set is among them.  If not, set to auto.  If set to auto, shift the first element from available types.


nao-pon:
@billynoah Thank you for fixing a bug. 👍

PR #3483, opened 2022-04-29
Update elFinderVolumeSFTPphpseclib.class.php

sdwru:
Currently, this class does not work with symlinks at all.  Permissions also don't work properly.  Using Flysystem is not an option for me because that requires PHP v8.  Older flysystem requires PHP v7.4 which is also not an option for me since I am stuck with v7.3 for now.  

I ended up creating my own `SFTP3phpseclib.class.php`.  I have edited this file with some of the changes I made to get symlinks sort of working.  I am still testing it so this is just a start.  It took me a long time to figure out how to get symlinks to show up so maybe this will help others struggling with this.  I am using phpseclib v3 so I only tested these changes with that library.

I also made a bunch of changes to get permissions working properly.  I may include that in a separate pull request after I test it more.

### DIFF
--- a/php/elFinderVolumeSFTPphpseclib.class.php
+++ b/php/elFinderVolumeSFTPphpseclib.class.php
@@ -247,14 +247,13 @@ class elFinderVolumeSFTPphpseclib extends elFinderVolumeFTP {
 
         $name = $info['filename'];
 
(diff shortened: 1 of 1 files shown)


nao-pon:
@sdwru thank you! I'm sorry to be late.

PR #3488, opened 2022-05-24
Fix trad

KiboOst:


nao-pon:
@KiboOst thank you! I'm sorry to be late.

PR #3499, opened 2022-07-28
Update mime.types to allow MS outlook message files

media253:
Adding application/vnd.ms-outlook mimetype with .msg extension


nao-pon:
@media253 Thanks! 👍

PR #3556, opened 2023-02-10
Update elFinderVolumeSFTPphpseclib.class.php

neuza-git:
this extra parenthesis on the if is causing syntax error


nao-pon:
@neuza-git Thanks! 👍

PR #3568, opened 2023-05-10
elFinder.js: Fix multiple calls to elFinder.sync() method when dispat…

magik092:
If the user has browser extensions installed that send a Message then this message is intercepted by js/elFinder.js:4827, and since the message is not a string of type JSON an exception is returned and the self.sync() method is called.
Using the example of the [wappalyzer](https://www.wappalyzer.com/) add-on - a dozen notifications are sent at once by which elFinder sends multiple requests at once to the application backend.

Here is an example for 32 queries through the wappalyzer add-on:
image

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -4821,7 +4821,7 @@ var elFinder = function(elm, opts, bootCallback) {
 		});
 
 		// bind window onmessage for CORS
-		$(window).on('message.' + namespace, function(e){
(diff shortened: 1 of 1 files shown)


nao-pon:
@magik092 Thank you for your suggestion. It seems better to deal with this problem in elFinder.sync(). I opened a related #3572, so I'm going to close this one.

Thank you again! 👍

PR #3573, opened 2023-06-02
Update elfinder.zh_TW.js

cssf998811:


nao-pon:
@cssf998811 Thanks! 👍

PR #3584, opened 2023-07-24
Update fonts.css

bilalkhan13:
Applied the dry methodology & sort the classes according to the font size


nao-pon:
@bilalkhan13 I'm sorry to be late. Merge.

PR #3585, opened 2023-07-24
 Eliminate the unnecessary data variable.

bilalkhan13:
Eliminate the unnecessary data variable and directly access the properties from self.data.


nao-pon:
@bilalkhan13 I'm sorry to be late. Merge.

PR #3586, opened 2023-07-24
Used switch statement instead of if-else

bilalkhan13:
Using a switch statement makes the code more readable and easier to manage when dealing with different cases based on the type.

### DIFF
--- a/js/worker/quicklook.unzip.js
+++ b/js/worker/quicklook.unzip.js
@@ -44,12 +44,22 @@ var type = self.data.type,
 	};
 
 self.res = {};
-if (type === 'tar') {
-	self.res.files = tarFiles(new Uint8Array(bin));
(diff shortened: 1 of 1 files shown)


js/worker/quicklook.unzip.js
```diff
@@ -44,12 +44,22 @@ var type = self.data.type,
...
+    self.res.files = unzipFiles.call(new Zlib.Unzip(bin));
+    break;
+  case 'gzip':
+    self.res.files = tarFiles(new Zlib.Gunzip(bin).decompress());
+    break;
+  case 'bzip2':
+    self.res.files = tarFiles(self.bzip2.simple(self.bzip2.array(bin)));
+    break;
+  default:
+
```
ManuMnesh, line 62:
```suggestion
    return 0;
```

PR #3604, opened 2023-10-10
MySQL driver: Use prepared statements instead of escaping when saving file

adamvalalsky:
Right now, inserting empty binary file (all binary zeroes) causes escaping of every single byte in content. This creates SQL which is double in size and can easily exceed max_allowed_packet value in MySQL.

This change uses prepared statements instead of escaping, so created SQLs (for files that would need escaping) are smaller.


nao-pon:
@adamvalalsky I'm sorry to be late. Merge.

PR #3607, opened 2023-10-17
Fixed elfinder.src.html in README instructions

filipopo:
I see that the file is actually elfinder.html and that it's correctly referenced in the wiki


nao-pon:
@filipopo `elfinder.src.html` is used when using raw source code without compressed js or css, so there is no problem leaving it as is. Thanks!

PR #3619, opened 2023-12-04
Update elFinder.class.php

Nounours13:
When you download a file via context menu the windows download popup don't close


nao-pon:
@Nounours13 It certainly seems that way. Thanks!

PR #3673, opened 2024-09-12
Accept download files URL without filter on local environment

aktolu:


nao-pon:
@aktolu This PR is expected to be vulnerable to HTTP header injection attacks. Can you ensure it is secure against such attacks?

aktolu:
> @aktolu This PR is expected to be vulnerable to HTTP header injection attacks. Can you ensure it is secure against such attacks?

I don't think there will be any problem. If HTTP_HOST is sent as fake, Apache will not run the script due to host incompatibility.

nao-pon:
Thank you for the PR, and sorry this sat for so long.

After review, I don’t think we can merge this change.

The issue is that it makes SSRF protection depend on `$_SERVER['HTTP_HOST']`, which comes from the request and should not be used as a trusted signal for changing security behavior. Depending on the server configuration, that could weaken the existing checks for loopback, private, and link-local addresses.

Also, the localhost detection is incomplete in practice, since values like `localhost:8000` would not match.

For those reasons, I’m closing this PR. A safer approach would be an explicit configuration option if this behavior ever needs to be supported.

PR #3677, opened 2024-10-11
Allow image URL in theme manifest.json to be a relative link as well

blutorange:
Relative links are resolved against the manifest.json, the same way as is done for the CSS URL.

This is needed when you are living places of the world where you need to host everything yourself, as linking to external servers is difficult due to privacy concerns. If your software can be hosted on different servers, everything needs to be relative to work properly.


nao-pon:
@blutorange Thanks!👍

PR #3682, opened 2024-11-05
Update elfinder.ru.js

Ruslan-Aleev:
Fixed incorrect translation of "Prefence / Prefences"


nao-pon:
@Ruslan-Aleev Thanks!👍

PR #3690, opened 2024-12-13
Fixes #3689, fm.sync removes unavailable volumes.

### LINKED ISSUE #3689: Refresh does not remove volumes not available anymore
**Describe the bug**

Volumes may stop being available (such as virtual volumes from a database and the volume was deleted from the database by another user). Doing a reload should detect this case and remove the volume, but currently, it does not.

**To Reproduce**

Open ElFinder with several volumes. Do something on the server that results in one volume not being available anymore. `Right click -> Reload` or call `fm.sync` on the file manager instance.

**Expected behavior**

ElFinder makes an `open` request to server, with `tree = 1`. The server replies with a list of all volumes.

If a volume is missing in the response from the server, ElFinder should remove that volume from the UI.

**Additional context**

ElFinder already checks for removed files / volumes when `fm.sync` is called. However, there's this part:

https://github.com/Studio-42/elFinder/blob/7544918a2ba656a4fe72a6fc8bb502d39f601cda/js/elFinder.js#L2507-L2527

Where ElFinder merged the roots it already knows about into the response. If I understand it correctly, it should not do that for the open command with `tree = 1`, as the server will already return all available volumes.

Adding a simple check so that it only calls `pushLeafRoots` when `tree != 1` seems to solve the issue -- the volume gets removed.

Here's a demonstration where another user removes a volume. In the first request, the server still returns the volume `New theme 4`. After it was deleted, the server does not return it anymore. After adding the above-mentioned check, refreshing removes the volume: 

[Screencast from 2024-12-13 20-31-25.webm](https://github.com/user-attachments/assets/ebe2fd75-868f-4b48-ae7d-c5010eda6651)

blutorange:
A potential solution that fixes #3689

When a request is made to the server with the open command and tree=1, the server must reply with a list of all volumes. If the server's response does not include a volume ElFinder already knows about, that volume should be treated as not available anymore and get removed from the file explorer UI.

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -3094,9 +3094,11 @@ var elFinder = function(elm, opts, bootCallback) {
 				return c;
 			},
 			comp  = compare(),
+			odataRoots,
 			dfrd  = $.Deferred().always(function() { !reqFail && self.trigger('sync'); }),
+			tree = (! onlydir && this.ui.tree) ? 1 : 0,
 			opts = [this.request({
-				data           : {cmd : 'open', reload : 1, target : cwd, tree : (! onlydir && this.ui.tree) ? 1 : 0, compare : comp},
+				data           : {cmd : 'open', reload : 1, target : cwd, tree : tree, compare : comp},
 				preventDefault : true
 			})],
 			exParents = function() {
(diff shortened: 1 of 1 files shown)


blutorange:
Hmm, that only seems to work in the root directory. When using the refresh command in a subfolder, elFinder makes sends an `tree` command where it also merges the root volumes.

blutorange:
49b27aaa8660343e2746e44876341981b78b8c1b is an amendment to the previous commit. When invoking the refresh action on a non-root folder, ElFinder also makes a "tree" request to the server and adds all root volumes it currently knows about. Since other parts of the code might rely upon this behavior (such as `tree.js`), remove
non-existent volumes from the response to the tree command in the `fm.sync` method.

PR #3735, opened 2025-10-18
Fix image editor save error after changing directories

jmario74:
when file is opened in the current directory, store the file node with "fm.file" function; then use it if the "fm.file" function can not find the file in the switched directory.


nao-pon:
Thanks for the PR.

I can see the intention here, but I’m not comfortable merging this as-is.

I understand the reported issue in #3728, but on my side I could not clearly reproduce a save failure just by switching directories. After saving, elFinder posts the content back to the original file hash with `put`, and `navigate.target = 'changed'` then moves focus back to the changed file, so returning to the original directory after save looks like the expected existing behavior rather than the bug itself. ([[GitHub](https://github.com/Studio-42/elFinder/issues/3728)][1])

Because of that, I’m not yet convinced that storing an extra `loadedFile` object on the editor is the right fix. In the current code, the MIME is already kept in the editor/dialog state (`$base.data('mime')` in the TUI path, and `ifm.data('mime')` in the Photopea path), so if `fm.file(hash)` can sometimes become unavailable, this looks like it should be handled with a smaller guard that simply preserves the existing MIME instead of introducing additional state. ([[GitHub](https://github.com/Studio-42/elFinder/blob/master/js/extras/editors.default.js)][2])


For now, I’m going to leave this unmerged unless we can get a clearer reproduction path or a smaller fix.

[1]: https://github.com/Studio-42/elFinder/issues/3728 "[BUG] TUIImageEditor stops work when switching to a different working directory · Issue #3728 · Studio-42/elFinder · GitHub"
[2]: https://github.com/Studio-42/elFinder/blob/master/js/extras/editors.default.js "elFinder/js/extras/editors.default.js at master · Studio-42/elFinder · GitHub"

jmario74:
Thanks nao-pon for taking a look at this, i have revised the fix using your guidance and used "this.file.mime" that has the stored file mime and supplied it to "$base.data('mime', this.file.mime)"; both "TUI" & "Photopea" are working properly. this looks like an isolated case, the bug was reported a while ago and there has been no additional reports since then; so i think this can be closed now and can serve as a reference to the issue for anybody who encounters it.

PR #3737, opened 2025-10-18
pressing minimized button ( - ) on the editor while the editor is maximized will close the editor

jmario74:
when the editor is minimized from a maximized state it still has the class "elfinder-maximized" which causes it to close, removing the "elfinder-maximized" class on minimize prevents it from closing.

MAXIMIZED:
max

MINIMIZED:
Screenshot from 2025-10-18 12-51-54

### DIFF
--- a/js/ui/dialog.js
+++ b/js/ui/dialog.js
@@ -277,7 +277,7 @@ $.fn.elfinderdialog = function(opts, fm) {
 									}).position();
 									mnode = dialog.clone().on('mousedown', function() {
 										$this.trigger('mousedown');
-									}).removeClass('ui-draggable ui-resizable elfinder-frontmost');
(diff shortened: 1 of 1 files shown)


nao-pon:
@jmario74 Thanks for the PR.

I can confirm that the `elfinder-maximized` class remains on the minimized dialog clone, so removing it makes sense as a cleanup.

However, from my testing, the dialog does not actually close. It is minimized to the title bar, and clicking the minimized title bar restores it again. So it seems possible that this was interpreted as a close behavior, while it is actually a minimized-state inconsistency.

I’m OK with merging this cleanup, but I don’t think it fully explains or resolves a real close issue.

If you have exact reproduction steps for a case where the dialog is truly closed rather than minimized, please share them.

PR #3738, opened 2025-10-18
added "SVG image" & "WEBP image" file type description to "Kind" column when in list view

jmario74:
BEFORE:
before
AFTER:
after


nao-pon:
@jmario74 Thanks! 👍

PR #3739, opened 2025-10-26
Update deprecated php type casts

colemanw:
In PHP 8.5, non-canonical scalar type casts (boolean|double|integer|binary) deprecated.

See https://php.watch/versions/8.5/boolean-double-integer-binary-casts-deprecated.

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -4394,7 +4394,7 @@ var go = function() {
     protected function utime()
     {
         $time = explode(" ", microtime());
-        return (double)$time[1] + (double)$time[0];
+        return (float)$time[1] + (float)$time[0];
     }
 
     /**


escopecz (approved):
We are also getting the deprecation. This is the suggested solution:

```
vendor/studio-42/elfinder/php/elFinder.class.php:4392
Non-canonical cast (double) is deprecated, use the (float) cast instead
```

escopecz:
@troex do you think this could be merged and released? It's blocking support for PHP 8.5 for Mautic. All other dependencies are supporting PHP 8.5 now.

escopecz:
Oh man, it would be ridiculous to fork this Composer Package just because of this one line change, wouldn't it?

nao-pon:
@colemanw Thanks for this cleanup.

Using `(float)` instead of `(double)` is fully compatible with older PHP versions, since `(double)` is just an alias. Switching to the canonical form also makes sense for forward compatibility with newer PHP versions where non-canonical casts are deprecated.

This looks good to merge.

colemanw:
Thanks for merging @nao-pon. It would be great to put out a new release so that ElFinder is compatible with php 8.5.

PR #3561, opened 2023-03-22
Correctly urlencode path in setcookie(); fix #3538

ctrlaltca:
Some characters are valid in URLs but not in cookies, eg. comma or semicolon.
They needs to be encoded properly.

Issue #3538, opened 2022-11-14
Php error while trying to download file related to setcookie()


ctrlaltca:
**Describe the bug**
I try to download a file using elFinder clicking over the "floppy disk" icon.
I receive a file, but instead of the actual file contents it contains an html page with the following php error:

```
ValueError
Description
setcookie(): "path" option cannot contain ",", ";", " ", "\t", "\r", "\n", "\013", or "\014"
Source File
/<hidden>/vendor/studio-42/elfinder/php/elFinder.class.php (2085)
2074:             $ua = isset($_SERVER['HTTP_USER_AGENT'])? $_SERVER['HTTP_USER_AGENT'] : '';
2075:             if (preg_match('/MSIE [4-8]/', $ua)) { // IE < 9 do not support RFC 6266 (RFC 2231/RFC 5987)
2076:                 $filename = 'filename="' . $filenameEncoded . '"';
2077:             } elseif (strpos($ua, 'Chrome') === false && strpos($ua, 'Safari') !== false && preg_match('#Version/[3-5]#', $ua)) { // Safari < 6
2078:                 $filename = 'filename="' . str_replace('"', '', $file['name']) . '"';
2079:             } else { // RFC 6266 (RFC 2231/RFC 5987)
2080:                 $filename = 'filename*=UTF-8\'\'' . $filenameEncoded;
2081:             }
2082:         }
2083: 
2084:         if ($args['cpath'] && $args['reqid']) {
2085:             setcookie('elfdl' . $args['reqid'], '1', 0, $args['cpath']);
2086:         }
2087: 
2088:         $result = array(
2089:             'volume' => $volume,
2090:             'pointer' => $fp,
2091:             'info' => $file,
2092:             'header' => array(
2093:                 'Content-Type: ' . $mime,
2094:                 'Content-Disposition: ' . $disp . '; ' . $filename,
2095:                 'Content-Transfer-Encoding: binary',
2096:                 'Content-Length: ' . $file['size'],
2097:                 'Last-Modified: ' . gmdate('D, d M Y H:i:s T', $file['ts']),

Stack Trace
#0 /<hidden>/vendor/studio-42/elfinder/php/elFinder.class.php(2085): setcookie()
#1 /<hidden>/vendor/studio-42/elfinder/php/elFinder.class.php(1210): elFinder->file()
#2 /<hidden>/vendor/studio-42/elfinder/php/elFinderConnector.class.php(160): elFinder->exec()
[...]

Issue #3528, opened 2022-10-19
Download not working for Windows Server


tru-wp-dev:
**Describe the bug**
Trying to download a folder in my local system 
OS - windows 

Download is happening with Failed - No File like the below screenshot
https://prnt.sc/6knWaVVFdd1S

PR #3754, opened 2026-01-12
Correctly urlencode path in setcookie(); refix Studio-42#3538

ctrlaltca:
Some characters are valid in URLs but not in cookies, eg. comma or semicolon. They needs to be encoded properly.

```
setcookie(): "path" option cannot contain ",", ";", " ", "\t", "\r", "\n", "\013", or "\014" (Error Code: 0)
vendor/studio-42/elfinder/php/elFinder.class.php (2085)
```

This is the same as issue #3538 
It was fixed by #3561 
Then the change was reverted in #3619 
I'm not sure about the problem fixed by #3619, as there's no related issue or explanation on the change.

Still, the original problem of #3528 reappeared, so I'm submitting the fix again.

### DIFF
--- a/php/elFinder.class.php
+++ b/php/elFinder.class.php
@@ -2087,7 +2087,7 @@ class elFinder
         }
 
         if ($args['cpath'] && $args['reqid']) {
-            setcookie('elfdl' . $args['reqid'], '1', 0, $args['cpath']);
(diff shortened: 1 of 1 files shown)


nao-pon:
@ctrlaltca You're absolutely right. Thank you! 👍

PR #3757, opened 2026-02-01
Update Russian translation and authorship details

alexantr:


nao-pon:
@alexantr Thanks! :+1:

PR #3759, opened 2026-02-14
Fix incorrect URL generation in getContentUrl()

### LINKED ISSUE #3746: elfinder cannot open the file that has just been uploaded
I have downloaded the new version of elfinder 2.1.66 and am trying to integrate it. I found some mistake. if upload a file (any text file, for example) and double-click on it immediately after uploading, a new browser window will open for viewing with an error message that the file was not found. 
I see that the file URL is incorrect.
For example, I'm trying to open a file https://sile.com/files/test.txt
but after uploading and just opening I see wrong url
On my local computer: http://sile.com/files//test.txt
(note the strange double slash "files//test.txt ")
And it's the same on my hosting, but the URL looks like
https://sile.com/files/s/test.txt
(note the strange addition of the string "/s").
This is not a real problem, because everything works correctly after the page is reloaded, but I think this bug needs to be fixed.

darth-antony:
trim($this->root, '/') strips the leading slash from the root path, making strlen() return a value 1 less than expected. This causes substr() to include an extra character from the root folder name in the relative path.

For example, with root folder "images":
- Expected URL: /files/image.jpg
- Actual URL:   /files/s/image.jpg

Using strlen($this->root) + 1 is consistent with other URL construction methods in elFinderVolumeLocalFileSystem (lines 206, 612, 845, 1420) which all use this correct pattern.

Fixes #3746

### DIFF
--- a/php/elFinderVolumeDriver.class.php
+++ b/php/elFinderVolumeDriver.class.php
@@ -3331,7 +3331,7 @@ abstract class elFinderVolumeDriver
             }
         }
         if (empty($file['url']) && $this->URL) {
-            $path = str_replace($this->separator, '/', substr($this->decode($hash), strlen(trim($this->root, '/' . $this->separator))));
+            $path = str_replace($this->separator, '/', substr($this->decode($hash), strlen($this->root) + 1));
             if ($this->encoding) {
                 $path = $this->convEncIn($path, true);
             }


Smanst3r:
Any news on this https://github.com/Studio-42/elFinder/blob/master/php/elFinderVolumeDriver.class.php#L3334?
I have the same issue but I thought it could be done with `rtrim($this->root..)...` instead of `trim($this->root...`
and return `rtrim($this->URL, '/') . $path;`
Kind regards

nao-pon:
@darth-antony That makes sense.

I think this became visible mainly in the immediate post-upload flow, where the backend has not yet exposed a stable stat URL and `getContentUrl()` is used directly. After reload, the normal stat data path hides the issue.

So the root cause explanation in the PR may be a bit simplified, but the fix direction itself still looks valid.

PR #3761, opened 2026-02-27
Fixing bug where uploading large files sometimes fails

andresrs:
Sometimes uploading large files fails with the following error: Unsupported operand types: string + int {"userId":2,"exception":"[object] (TypeError(code: 0): Unsupported operand types: string + int at <path>/vendor/studio-42/elfinder/php/elFinder.php:4476). 

The function file_get_contents returns a string, and PHP 8 doesn't do implicit conversion. It requires the conversion to be explicit to not give problems.


nao-pon:
@andresrs Thanks for the fix and the clear explanation.

This looks like a real PHP 8 compatibility bug with a practical impact on uploads.
The change is small and the failure mode is well described, so this is a good candidate for merging after a quick verification.

PR #3762, opened 2026-03-01
Refactor touch event binding in elfinder.js

mulgaBrian:
for jquery 4.0


nao-pon:
@mulgaBrian Thanks again for your jQuery 4 related work.
I’ve retargeted this PR to jquery4 so we can review and test the migration in a dedicated branch first.

PR #3771, opened 2026-03-01
Use Array.isArray instead of $.isArray

mulgaBrian:
for jquery 4.0


colemanw:
This is a no-brainer to merge @nao-pon

PR #3797, opened 2026-07-29
Fix postMessage origin bypass and playsound DOM XSS

wpexpertsio:
## Summary
- Replace prefix `indexOf(origin)` with exact `URL.origin` matching
- Prevent playsound DOM XSS via filename allowlist + DOM APIs
- Sanitize untrusted `bind` from postMessage payloads

## Test plan
- [ ] Same-origin / connector-origin postMessage still works
- [ ] Prefix attacker origin is rejected
- [ ] Malicious `soundFile` does not inject HTML / XSS
- [ ] Normal `rm.wav` still plays when sound is enabled

## Credit
WPExperts / Advanced File Manager

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -4979,9 +4979,6 @@ var elFinder = function(elm, opts, bootCallback) {
 			try {
 				trusted[new URL(self.convAbsUrl(self.uploadURL)).origin] = true;
 			} catch (ignore) {}
-			try {
-				trusted[window.location.origin] = true;
-			} catch (ignore) {}
 			if (res && res.origin && trusted[res.origin]) {
 				try {
 					try {
@@ -4998,7 +4995,7 @@ var elFinder = function(elm, opts, bootCallback) {
 					} 
 					if (data) {
 						bind = obj.bind;
-						if (bind && !/^[a-zA-Z0-9;._-]+$/.test(bind)) {
+						if (bind && !/^[a-zA-Z0-9._-]+$/.test(bind)) {
 							bind = '';
 						}
 						if (data.error) {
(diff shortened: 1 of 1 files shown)


nao-pon:
@wpexpertsio

Thanks for the PR! The changes generally look good to me.

I have two questions before reviewing further:

1. I noticed that `window.location.origin` was added to the trusted origins list. Could you explain the reasoning behind trusting the current page's origin in addition to the connector and upload URL origins? I'm wondering if there was a specific scenario this is intended to support.

2. I also noticed that the validation regex for `bind` allows a semicolon (`;`):

```js
/^[A-Za-z0-9;._-]+$/
```

Is there a particular reason for allowing `;`? Are there any existing event names that require it, or would `[A-Za-z0-9._-]` be sufficient?

Thanks!

wpexpertsio:
Thanks for the review!

`window.location.origin`
There wasn’t a strong requirement for it for this security fix. We included it as a convenience for same-page hosts, but the connector/uploadURL origins are enough. We’ve removed it, so we only trust those exact origins.

Semicolon in the bind regex
No existing event name needs `;`. It was overly permissive. We’ve tightened the pattern to `/^[A-Za-z0-9._-]+$/`.

Pushed the latest commit on this branch. Please take another look - thanks!

nao-pon:
@wpexpertsio Thanks! 👍

nao-pon:
@wpexpertsio

Thank you for the fix. I accidentally merged this PR into 2.1-src, but the change should originally be applied to the source files in master, since js/elFinder.js in 2.1-src is generated by the build process.

The file to be modified is js/elfinder.js.

Could you please open a new PR targeting master and apply the same fixes to the corresponding source files? Sorry for the confusion and thank you for your help.

PR #3798, opened 2026-08-03
Fix postMessage origin bypass and playsound DOM XSS (master source)

wpexpertsio:
## Summary
- Apply the same security fixes from #3797 to the master source file `js/elFinder.js`
- Replace prefix `indexOf(origin)` with exact `URL.origin` matching
- Prevent playsound DOM XSS via filename allowlist + DOM APIs
- Sanitize untrusted `bind` from postMessage payloads

Follow-up to #3797 as requested: the previous change landed on `2.1-src` (generated build), so this PR targets `master` source.

## Test plan
- [ ] Same-origin / connector-origin postMessage still works
- [ ] Prefix attacker origin is rejected
- [ ] Malicious `soundFile` does not inject HTML / XSS
- [ ] Normal `rm.wav` still plays when sound is enabled

## Credit
WPExperts / Advanced File Manager

### DIFF
--- a/js/elFinder.js
+++ b/js/elFinder.js
@@ -4714,9 +4714,22 @@ var elFinder = function(elm, opts, bootCallback) {
 	if (true === this.options.sound) {
 		this.bind('playsound', function(e) {
 			var play  = beeper.canPlayType && beeper.canPlayType('audio/wav; codecs="1"'),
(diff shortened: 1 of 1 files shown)


nao-pon:
@wpexpertsio Thanks for creating a new PR against `master`, and sorry for the extra work!

